PR #2, opened 2024-01-24
Pfs 3 header and footer

Nicholassully:


internal/model/assignee.go
```diff
@@ -0,0 +1,33 @@
...
+	Name        string   `json:"displayName"`
+	PhoneNumber string   `json:"phoneNumber"`
+	Deleted     bool     `json:"deleted"`
+	Email       string   `json:"email"`
+	Firstname   string   `json:"firstname"`
+	Surname     string   `json:"surname"`
+	Roles       []string `json:"roles"`
+	Locked      bool     `json:"locked"`
+	Suspended   bool     `json:"suspended"`
+}
```
denialanderror, line 19: Do we need all this assignee information? We should only be adding to the struct the data we need, otherwise it all bloats unnecessarily.

internal/server/environment_vars.go
```diff
@@ -17,7 +17,7 @@ func NewEnvironmentVars() (EnvironmentVars, error) {
 	return EnvironmentVars{
 		Port:            getEnv("PORT", "1234"),
 		WebDir:          getEnv("WEB_DIR", "web"),
-		SiriusURL:       getEnv("SIRIUS_URL", "http://localhost:8080"),
+		SiriusURL:       getEnv("SIRIUS_URL", "http://host.docker.internal:8080"),
```
denialanderror, line 20: It would be better to pass this through in a compose file, rather than default it here

web/template/inserts/header.gotmpl
```diff
@@ -0,0 +1,3 @@
+{{ define "header" }}
+  <sirius-header {{ if eq .Error "" }} {{ end }}></sirius-header>
```
denialanderror, line 2: What's the point of this if condition?

denialanderror (changes requested):

internal/model/assignee.go
Nicholassully: done

internal/server/environment_vars.go
Nicholassully: done

web/template/inserts/header.gotmpl
Nicholassully: done

denialanderror (commented):
Why are user details being fetched in this PR if they are never being used?

Nicholassully:
They dont need to fetch in this one but they are need in pfs-5 so should of been done in that one. For now is it ok if they go in as I will need them for the next ticket I am doing?

denialanderror (approved):

PR #3, opened 2024-01-26
Pfs 4 adding tests

Nicholassully:


cypress/e2e/accessibility.cy.js
```diff
@@ -0,0 +1,13 @@
+import "cypress-axe";
+
+describe("Accessibility test client tasks", { tags: "@axe" }, () => {
```
josephsmith0705, line 3: I might be mistaken, but should this definitely be 'client tasks', not finance hub?

josephsmith0705 (dismissed):

cypress/e2e/accessibility.cy.js
Nicholassully: yes, yes it should be I will change it :D

josephsmith0705 (approved):

PR #5, opened 2024-02-08
pfs-9 adding client information #minor

Nicholassully:


cypress/e2e/finance_hub.cy.js
```diff
@@ -0,0 +1,25 @@
+describe("Finance Hub", () => {
+    beforeEach(() => {
+        cy.setCookie("Other", "other");
+        cy.setCookie("XSRF-TOKEN", "abcde");
+        cy.visit("/supervision/finance/2");
+    });
+
+    describe("Finance Details Header", () => {
+        it("should show the finance person name", () => {
+            cy.get('#hook-person-name').should("contain", "Finance Person");
```
denialanderror, line 10: You should use `data-cy` attributes instead of ids if that is how you are wanting to address elements. ids are for CSS and JS, not testing.

Even better, don't address elements by id/attribute at all and do it how a user would: by text.

cypress/e2e/finance_hub.cy.js
```diff
@@ -0,0 +1,25 @@
+describe("Finance Hub", () => {
+    beforeEach(() => {
+        cy.setCookie("Other", "other");
+        cy.setCookie("XSRF-TOKEN", "abcde");
```
denialanderror, line 4: We don't need an xsrf token here. There are no forms being submitted.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/opg-sirius-finance-hub
 go 1.21
 
 require (
+	github.com/gorilla/mux v1.8.1
```
denialanderror, line 6: What's mux being used for? This is a library that was deprecated and then undeprecated. If we want a web framework over the native http library, there are better, more stable options.

internal/server/finance_vars.go
```diff
@@ -1,15 +1,19 @@
...
+	"golang.org/x/sync/errgroup"
 	"net/http"
+	"strconv"
 )
 
 type FinanceVars struct {
 	Path            string
 	XSRFToken       string
 	MyDetails       model.Assignee
+	FinancePerson   model.FinancePerson
```
denialanderror, line 16: This can just be called "Person", or better, "Client". There isn't any other type of person here.

json-server/config/routes.json
```diff
@@ -1,4 +1,5 @@
 {
     "/api/v1/*": "/$1",
-    "/users/current": "/users-current"
+    "/users/current": "/users-current",
+    "/finance/:id": "/finance"
```
denialanderror, line 4: You don't need to put in custom routing here. json-server will take a route and look in the data structure for the array matching that path and the object with that id. Structure the data better rather than adding in custom routes for everything, otherwise the db.json gets very messy, very quickly. https://github.com/typicode/json-server

json-server/config/db.js
denialanderror, line 1: This should be a json file unless there is a good reason for it needing javascript

web/assets/scss/finance-hub.scss
denialanderror, line 1: Is this the same for all our hubs? If so, we should look to extract it as a package somewhere (not as part of this, we'll create a ticket).

web/template/invoices.gotmpl
denialanderror, line 1: These templates should have type annotations

internal/server/finance_vars_test.go
denialanderror, line 1: Why was this test deleted?

internal/sirius/get_finance_person_details.go
denialanderror, line 1: There doesn't appear to be a test for this

denialanderror (changes requested):

cypress/e2e/finance_hub.cy.js
Nicholassully: done

cypress/e2e/finance_hub.cy.js
Nicholassully: done

go.mod
Nicholassully: removed

internal/server/finance_vars.go
Nicholassully: done

json-server/config/routes.json
Nicholassully: done

json-server/config/db.js
Nicholassully: done

web/assets/scss/finance-hub.scss
Nicholassully: Yes it looks like this is the same in deputy and firm hubs

web/template/invoices.gotmpl
Nicholassully: done

internal/server/finance_vars_test.go
Nicholassully: It was by mistake

internal/sirius/get_finance_person_details.go
Nicholassully: added

web/template/layout/page.gotmpl
```diff
@@ -20,12 +20,12 @@
       <link href="{{ prefix "/stylesheets/all.css" }}" rel="stylesheet">
     </head>
 
-    <body class="govuk-template__body sirius-finance" data-test="My value" data-other="report:details">
+    <body class="govuk-template__body sirius-finance">
 
       {{ template "header" . }}
 
-      <div class="govuk-width-container app-width-container">
-        <main class="govuk-main-wrapper app-main-class" id="main-content" role="main">
```
josephsmith0705, line 28: Just FYI, we use this id="main-content" usually to include a skip link for accessibility purposes. It's not a huge deal removing this here as we haven't got a skip link in this system yet anyway, but all gov.uk sites are meant to have one so might be worth keeping this in if we're expecting to add it in before this goes live anyway.
https://design-system.service.gov.uk/components/skip-link/

josephsmith0705 (approved):

PR #6, opened 2024-02-09
Pfs 11 invoice tab

Nicholassully:


denialanderror (changes requested):

internal/server/invoices.go
denialanderror, line 1: There's no tests for this file and the code coverage is moaning. Is that because it isn't doing anything yet or was it missed? No probs if the former, we'll just need to remember to add them in later.

Nicholassully: Yeah at the minute its not doing anything and when we use it of course we will test it

denialanderror (approved):

PR #14, opened 2024-02-23
pfs-19 and pfs-27 invoice with drop down frontend done #minor

Nicholassully:


cypress/e2e/invoices.cy.js
```diff
@@ -13,4 +13,26 @@ describe("Invoice Tab", () => {
             cy.contains('[data-cy="outstanding-balance"]', "Outstanding Balance");
         });
     });
+
+    describe("Invoices ledger allocations", () => {
```
denialanderror, line 17: These tests are just testing the headers, not the actual data. It's also just checking those elements are on the page, not that they are in the right locations.

internal/model/invoices.go
```diff
@@ -0,0 +1,69 @@
+package model
+
+type InvoiceList struct {
+	Invoices []Invoice `json:"invoices"`
```
denialanderror, line 4: The API would probably be cleaner if /invoices returned an array of invoices rather than an object with an invoices array field.

internal/model/invoices.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type Invoice struct {
+	Id                 int                `json:"id"`
+	Ref                string             `json:"ref"`
+	Status             string             `json:"status"`
+	Amount             string             `json:"amount"`
+	RaisedDate         string             `json:"raisedDate"`
+	Received           string             `json:"received"`
```
denialanderror, line 13: These should probably be dates. We want our API to be agnostic of how the data is being used, so having these dates as preformatted date strings would be an anti-pattern

internal/sirius/get_invoices.go
```diff
@@ -0,0 +1,61 @@
...
+			Received:           t.Received,
+			OutstandingBalance: t.OutstandingBalance,
+			Ledgers:            t.Ledgers,
+			SupervisionLevels:  t.SupervisionLevels,
+		}
+		invoices = append(invoices, invoice)
+	}
+
+	invoiceList.Invoices = invoices
+	//see returning just one odject not an array
```
denialanderror, line 59: Not sure I understand this comment or what the above iteration is doing. `invoiceList.Invoices` is an array of invoices, and you are iterating through the array, copying the elements, and reassigning the same data back to the invoices list.

web/assets/main.js
```diff
@@ -1,6 +1,23 @@
...
+    const summaries = document.getElementsByClassName("summary")
+    for (let summary of summaries) {
+        let summeryId = summary.id
+        summary.onclick = function () {
+            toggleSummary(summeryId);
+        }
+    }
+}
+
+function toggleSummary(summeryId) {
```
denialanderror, line 19: summary not "summery"

web/assets/main.js
```diff
@@ -1,6 +1,23 @@
...
 
+
 window.htmx = require('htmx.org');
 
 htmx.logAll();
+
+if (document.querySelector(".summary")) {
+    const summaries = document.getElementsByClassName("summary")
+    for (let summary of summaries) {
+        let summeryId = summary.id
```
denialanderror, line 12: Use const over let unless you need it to be mutable

web/template/tabs/invoices-list.gotmpl
```diff
@@ -12,16 +12,110 @@
...
-          <th scope="col" data-cy="outstanding-balance" class="govuk-table__header">Outstanding Balance</th>
-        </tr>
-      </thead>
-    </table>
-{{ end }}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full ">
+        {{ $length := len .Invoices.Invoices }}
+        {{ if eq $length 0 }}
+{{/*        <tr class="govuk-table__row">*/}}
```
denialanderror, line 19: Comments?

web/template/tabs/invoices-list.gotmpl
denialanderror, line 1: There's a lot of inline styles here. These should all be css classes

internal/server/invoices_test.go
denialanderror, line 1: No test for the success path

denialanderror (changes requested):

internal/model/invoices.go
Nicholassully: done

internal/model/invoices.go
Nicholassully: done

internal/sirius/get_invoices.go
Nicholassully: done

web/assets/main.js
Nicholassully: done

web/assets/main.js
Nicholassully: done

web/template/tabs/invoices-list.gotmpl
Nicholassully: done

web/template/tabs/invoices-list.gotmpl
Nicholassully: done

internal/server/invoices_test.go
Nicholassully: done

cypress/e2e/invoices.cy.js
Nicholassully: I was thinking that in the cypress test I would only test the headers as they are static and unit test would look more into the data?

PR #15, opened 2024-02-29
Pfs 28 adjust invoice

Nicholassully:


cypress/e2e/invoices.cy.js
```diff
@@ -1,16 +1,55 @@
 describe("Invoice Tab", () => {
-    beforeEach(() => {
-        cy.visit("/clients/1/invoices");
-    });
-
-    describe("Invoices", () => {
-        it("shows table header", () => {
+        it("shows table headers", () => {
```
denialanderror, line 2: Most of these tests are named incorrectly

cypress/e2e/update-invoice.cy.js
```diff
@@ -0,0 +1,25 @@
...
+
+    });
+
+    it("shows correct success message", () => {
+        cy.setCookie("success-route", "/invoices?clientId=1?");
+        cy.visit("/clients/1/invoices");
+        cy.get(':nth-child(1) > :nth-child(7) > .moj-button-menu > .moj-button-menu__wrapper > .govuk-button').click()
+        cy.get('#writeOff').check({force:true});
+        cy.get('.govuk-button').click()
+        cy.get('.moj-banner__message').contains("The write off is now waiting for approval")
```
denialanderror, line 23: Can we also have an assertion that it navigates correctly?

internal/model/invoice_type.go
```diff
@@ -0,0 +1,15 @@
+package model
+
+type InvoiceType struct {
+	Handle         string
+	Description    string
+	AmountRequired bool
+}
+
+var InvoiceTypes = []InvoiceType{
+	{Handle: "writeOff", Description: "Write off", AmountRequired: false},
```
denialanderror, line 10: We'll keep it like this for now but `Handle` and `Description` are Sirius-specific. This won't be needed if we are sending this to a separate backend.

internal/server/invoices_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"github.com/opg-sirius-finance-hub/internal/model"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestInvoice_error(t *testing.T) {
```
denialanderror, line 11: This doesn't test an error

denialanderror (changes requested):

denialanderror (approved):

PR #22, opened 2024-03-20
pfs-49 backend for client summary and db schema

Nicholassully:


json-server/config/routes.json
```diff
@@ -1,4 +1,5 @@
 {
+    "/clients/:id": "/header-details?clientId=:id",
```
denialanderror, line 2: Same with this route. This is coming from the API now

yarn.lock
denialanderror, line 1: Remove

denialanderror (changes requested):

PR #33, opened 2024-04-08
Pfs 51 supervision level backend

Nicholassully:


finance-api/internal/service/invoices.go
```diff
@@ -43,18 +27,58 @@ func (s *Service) GetInvoices(clientID int) (*shared.Invoices, error) {
...
+	}
+	return supervisionLevels
+}
+
+func getLedgerAllocations(s *Service, ctx context.Context, invoiceID int) []shared.Ledger {
+	var ledgerAllocations []shared.Ledger
+	totalOfLedgerAllocationsAmount = 0
+	ledgerAllocationsRawData, err := s.Store.GetLedgerAllocations(ctx, pgtype.Int4{Int32: int32(invoiceID), Valid: true})
+	if err != nil {
+		return nil
```
denialanderror, line 63: Shouldn't this return an error?

denialanderror (changes requested):

finance-api/internal/service/invoices.go
```diff
@@ -6,34 +6,18 @@ import (
 	"github.com/opg-sirius-finance-hub/shared"
 )
 
+var totalOfLedgerAllocationsAmount int
```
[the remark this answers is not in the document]
Nicholassully: If scope it any low I can use it in both the for loop on line 28 and in the method on line 74 in less I have missed a better way

finance-api/internal/service/invoices.go
Nicholassully: I am not sure how I would be able to return an error as the invoice struct for ledgers only expect an array of allocations, happy to talk about it tho

migrations/000002_invoice_changers.down.sql
denialanderror, line 1: These will all need to change to the new migration format as we had to move from Golang-Migrate to Goose due to the former being poorly maintained.

finance-api/internal/service/invoices.go
denialanderror: If the query is incorrect or returns invalid data, this would silently fail and we would have no indication it was wrong until it failed for users in production. If the code cannot return an error, it has to be refactored so it can. We can't just ignore errors because they are hard to write.

Functions in Go can return multiple types. The `Store.GetLedgerAllocations` function two lines above does exactly that.

finance-api/internal/service/invoices.go
```diff
@@ -0,0 +1,85 @@
...
+			TransactionType: i.Type,
+			Status:          "Applied",
+		}
+		ledgerAllocations = append(ledgerAllocations, ledgerAllocation)
+		totalOfLedgerAllocationsAmount = totalOfLedgerAllocationsAmount + int(i.Amount)
+	}
+	return ledgerAllocations
+}
+
+func CheckForAllocation(bankDate pgtype.Date, datetime pgtype.Timestamp) shared.Date {
```
denialanderror, line 79: This is an oddly named function. `CheckForAllocation` implies it is going to return a boolean. Also, this should be lower case unless you are wanting to use this function outside of the module.

finance-api/internal/service/invoices.go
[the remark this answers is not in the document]
denialanderror: If you need the value in more than one place, have one of them supply it to the other, or calculate it independently. Global variables should only be used for constants.

denialanderror (changes requested):

Makefile
```diff
@@ -30,8 +30,6 @@ scan: setup-directories
 dev-up:
 	docker compose run --rm yarn
 	docker compose -f docker-compose.yml -f docker/docker-compose.dev.yml build finance-hub finance-api
-	docker compose -f docker-compose.yml -f docker/docker-compose.dev.yml up finance-hub yarn json-server finance-api sqlc-gen sirius-db 
-	docker compose -f docker-compose.yml -f docker/docker-compose.dev.yml run finance-migration
```
denialanderror, line 34: These shouldn't be removed.

finance-api/internal/service/invoices.go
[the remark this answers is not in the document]
denialanderror: This still needs removing.

denialanderror (changes requested):

finance-api/internal/service/invoices.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return &invoices, nil
+}
+
+func getSupervisionLevels(s *Service, ctx context.Context, invoiceID int) []shared.SupervisionLevel {
+	var supervisionLevels []shared.SupervisionLevel
+	supervisionLevelsRawData, err := s.Store.GetSupervisionLevels(ctx, pgtype.Int4{Int32: int32(invoiceID), Valid: true})
+	if err != nil {
+		return nil
```
denialanderror, line 46: Don't silently fail. This should also return an error if an error occurs.

finance-hub/internal/server/invoices.go
```diff
@@ -22,7 +51,51 @@ func (h *InvoicesHandler) render(v AppVars, w http.ResponseWriter, r *http.Reque
...
 
-	data := &InvoiceTab{invoices, v}
+	data := &InvoiceTab{h.transform(invoices), v}
 	data.selectTab("invoices")
 	return h.execute(w, r, data)
 }
+
+func (h *InvoicesHandler) transform(in shared.Invoices) Invoices {
+	var out Invoices
+	for _, f := range in {
```
denialanderror, line 61: `f` isn't a useful variable name. I don't mind single letter variables but they should at least be relevant to what they represent.

finance-hub/internal/server/invoices.go
```diff
@@ -22,7 +51,51 @@ func (h *InvoicesHandler) render(v AppVars, w http.ResponseWriter, r *http.Reque
...
+			Ledgers:            transformLedgers(f.Ledgers),
+			SupervisionLevels:  transformSupervisionLevels(f.SupervisionLevels),
+		})
+	}
+	return out
+}
+
+func transformSupervisionLevels(in []shared.SupervisionLevel) []SupervisionLevel {
+	var out []SupervisionLevel
+	for _, f := range in {
```
denialanderror, line 79: Same here

finance-hub/internal/server/invoices.go
```diff
@@ -22,7 +51,51 @@ func (h *InvoicesHandler) render(v AppVars, w http.ResponseWriter, r *http.Reque
...
+			Received:           util.IntToDecimalString(f.Received),
+			OutstandingBalance: util.IntToDecimalString(f.OutstandingBalance),
+			Ledgers:            transformLedgers(f.Ledgers),
+			SupervisionLevels:  transformSupervisionLevels(f.SupervisionLevels),
+		})
+	}
+	return out
+}
+
+func transformSupervisionLevels(in []shared.SupervisionLevel) []SupervisionLevel {
```
denialanderror, line 77: These transform functions should be member functions of the handler, otherwise they will be in the module namespace.

migrations/20240410094040_ledger_allocation_type.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+ALTER TABLE ledger_allocation ADD COLUMN transaction_type varchar(250);
```
denialanderror, line 2: If you want to set varchar to its maximum (which presumably is what you are doing), you want varchar(255)

denialanderror:
Can you also take the data and routes out of json server, so there is no tidying up to do once we've done all the back end routes?

migrations/20240410094040_ledger_allocation_type.sql
Nicholassully: done

finance-hub/internal/server/invoices.go
Nicholassully: done

finance-hub/internal/server/invoices.go
Nicholassully: done

finance-api/internal/service/invoices.go
Nicholassully: done

migrations/20240410093925_invoice_changers.sql
denialanderror, line 1: This file should be renamed to detail what the migration does

finance-api/internal/service/invoices_test.go
```diff
@@ -0,0 +1,88 @@
...
+			s := &Service{
+				Store: Store,
+			}
+			got, err := s.GetInvoices(tt.id)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GetInvoices() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if (err == nil) && len(*got) == 0 {
+				assert.Empty(t, got)
```
denialanderror, line 80: This will assert as true if the test that is meant to return data returns nothing

migrations/20240410093925_invoice_changers.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+ALTER TABLE invoice ADD COLUMN createddate date;
+ALTER TABLE invoice ADD COLUMN createdby_id integer;
+
+-- +goose Down
+ALTER TABLE invoice DROP COLUMN person_id;
```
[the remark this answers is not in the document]
Nicholassully: They are columns that need to be dropped as will not be needed moving forward

finance-api/internal/service/invoices_test.go
Nicholassully: Need this for the second test as I can do a deep equal on an empty array. In less I have missed something

migrations/20240410093925_invoice_changers.sql
Nicholassully: done

finance-api/internal/service/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+	ctx := context.Background()
+
+	var invoices shared.Invoices
+
+	invoicesRawData, err := s.Store.GetInvoices(ctx, pgtype.Int4{Int32: int32(clientID), Valid: true})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, i := range invoicesRawData {
```
denialanderror, line 20: Rename `i` please. It isn't the index.

finance-api/internal/service/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+		ledgerAllocations, totalOfLedgerAllocationsAmount, err := getLedgerAllocations(s, ctx, int(i.ID))
+		if err != nil {
+			return nil, err
+		}
+		log.Println(ledgerAllocations)
+
+		supervisionLevels, err := getSupervisionLevels(s, ctx, int(i.ID))
+		log.Println(supervisionLevels)
```
denialanderror, line 28: multiple log statements

finance-api/internal/service/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func getSupervisionLevels(s *Service, ctx context.Context, invoiceID int) ([]shared.SupervisionLevel, error) {
+	var supervisionLevels []shared.SupervisionLevel
+	supervisionLevelsRawData, err := s.Store.GetSupervisionLevels(ctx, pgtype.Int4{Int32: int32(invoiceID), Valid: true})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, i := range supervisionLevelsRawData {
```
denialanderror, line 58: Same here

finance-api/internal/service/invoices.go
```diff
@@ -0,0 +1,97 @@
...
+
+func getLedgerAllocations(s *Service, ctx context.Context, invoiceID int) ([]shared.Ledger, int, error) {
+	var ledgerAllocations []shared.Ledger
+	totalOfLedgerAllocationsAmount := 0
+	ledgerAllocationsRawData, err := s.Store.GetLedgerAllocations(ctx, pgtype.Int4{Int32: int32(invoiceID), Valid: true})
+	if err != nil {
+		return nil, 0, err
+	}
+
+	for _, i := range ledgerAllocationsRawData {
```
denialanderror, line 78: And here

finance-api/internal/service/invoices_test.go
denialanderror: You'll need to think of a different way of doing it then. A test that passes on false positives isn't a test.

migrations/20240410093925_invoice_changers.sql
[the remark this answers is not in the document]
denialanderror: Why are you dropping them in the down migration then? This will only run if the migration is rolled back.

finance-api/internal/service/invoices.go
Nicholassully: done

finance-api/internal/service/invoices.go
Nicholassully: done

finance-api/internal/service/invoices_test.go
denialanderror: How does this fix the issue? The first test case expects an array containing data. If the function is broken and returns an array containing no data, this test will still pass.

migrations/20240410093925_invoice_created_fields.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE invoice ADD COLUMN createddate date;
+ALTER TABLE invoice ADD COLUMN createdby_id integer;
+
+-- +goose Down
+ALTER TABLE invoice DROP COLUMN person_id;
+ALTER TABLE invoice DROP COLUMN supervisionlevel;
+ALTER TABLE invoice DROP COLUMN confirmeddate;
+ALTER TABLE invoice DROP COLUMN batchnumber;
+ALTER TABLE invoice DROP COLUMN createddate;
+ALTER TABLE invoice DROP COLUMN createdby_id;
```
denialanderror, line 11: This still needs changing. Down migrations are for rolling back, not for dropping columns. If you want to drop unused columns as part of a migration, they need to go in the Up section

denialanderror (approved):

PR #51, opened 2024-05-02
pfs-74: JS fixes

denialanderror:
Fixes event listener registration issues by moving them inside an htmx.onLoad function block. Updates docker compose config so yarn watches and hotreloads as expected.

### DIFF
--- a/finance-hub/web/assets/main.js
+++ b/finance-hub/web/assets/main.js
@@ -1,4 +1,4 @@
-import { initAll } from 'govuk-frontend'
+import {initAll} from 'govuk-frontend'
 import "govuk-frontend/dist/govuk/all.mjs";
 import "opg-sirius-header/sirius-header.js";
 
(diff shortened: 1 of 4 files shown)


finance-hub/web/assets/main.js
```diff
@@ -9,43 +9,24 @@ window.htmx = require('htmx.org');
...
+    htmx.findAll(content, ".summary").forEach((element => {
+        htmx.on(`#${element.id}`, "click", () => htmx.toggleClass(htmx.find(`#${element.id}-reveal`), "hide"));
+    }));
 
-if (document.querySelector(".moj-banner--success")) {
-    const el = document.getElementsByClassName("moj-banner--success")[0];
-    el.onclick = () => el.classList.add("hide");
-}
+    htmx.findAll(".show-input-field").forEach((element) => {
+        element.addEventListener("click", () => htmx.removeClass(htmx.find("#field-input"), "hide"));
```
mattmachell, line 19: Question out of curiosity. Does HTMX do event delegation under the hood or bind only to the elements it initially finds via `findAll`? Might make this brittle to new DOM additions as it is currently written, though I don't know if that'll be an issue for this app.

jackgoodby (approved):

PR #57, opened 2024-05-16
pfs-22 front and back end with tests

Nicholassully:


finance-api/cmd/api/cancel_fee_reduction_test.go
```diff
@@ -0,0 +1,131 @@
...
+	server := Server{Service: mock, Validator: validator}
+	server.cancelFeeReduction(w, req)
+
+	res := w.Result()
+	defer res.Body.Close()
+	data, _ := io.ReadAll(res.Body)
+
+	expected := `
+{"Message":"","validation_errors":{"CancelFeeReductionNotes":{"thousand-character-limit":"This field CancelFeeReductionNotes needs to be looked at thousand-character-limit"}}}`
+
```
kate-49, line 106: Just checking, is this the real error message Sirius returns or just a fake one for the test?

finance-api/internal/service/fee_reductions.go
```diff
@@ -28,12 +28,20 @@ func (s *Service) GetFeeReductions(id int) (*shared.FeeReductions, error) {
...
+		feeReduction.FeeReductionCancelAction = showFeeReductionCancelBtn(feeReduction.Status)
 		feeReductions = append(feeReductions, feeReduction)
 	}
 
 	return &feeReductions, nil
 }
 
+func showFeeReductionCancelBtn(status any) bool {
+	if status == "Pending" || status == "Active" {
+		return true
```
kate-49, line 40: Do you have access to the ref data used for this status? Just wondering if you could add it as a const to make this a bit more foolproof?

finance-api/cmd/api/cancel_fee_reduction_test.go
Nicholassully: This is the error that we expect from the finance backend we are not going to Sirius for this call

kate-49: Ah ok, found the answer later on. This might be me being picky but is the naming convention camelcase for these names? I noticed the ones later on tend to be but I wasn't sure if you'd agreed on a set of rules for this.

finance-hub/internal/server/fee_reductions.go
```diff
@@ -36,22 +39,25 @@ func (h *FeeReductionsHandler) render(v AppVars, w http.ResponseWriter, r *http.
 	if err != nil {
 		return err
 	}
-
-	data := &FeeReductionsTab{h.transform(feeReductions), strconv.Itoa(ctx.ClientId), v}
+	clientIDString := strconv.Itoa(ctx.ClientId)
+	data := &FeeReductionsTab{h.transform(feeReductions, clientIDString), clientIDString, v}
```
kate-49, line 43: Is there a risk that this will cause an error?

finance-api/internal/service/fee_reductions.go
Nicholassully: We dont have ref data but making it a shared const is a great idea and moved to the shared package

finance-api/cmd/api/cancel_fee_reduction_test.go
Nicholassully: The key `CancelFeeReductionNotes` comes for the json object from this method
	if err != nil {
		errors := err.(validator.ValidationErrors)
		for _, fieldError := range errors {
			field := fieldError.Field()
			tag := fieldError.Tag()

			if description != "" {
				if validationErrors[description] == nil {
					validationErrors[description] = make(map[string]string)
				}
				validationErrors[description][tag] = fmt.Sprintf("This field %s needs to be looked at %s", description, tag)
			} else {
				if validationErrors[field] == nil {
					validationErrors[field] = make(map[string]string)
				}
				validationErrors[field][tag] = fmt.Sprintf("This field %s needs to be looked at %s", field, tag)
			}
		}

		return shared.ValidationError{
			Errors: validationErrors,
		}
	}

	return shared.ValidationError{}
}
Thats the reason for the casing. The others that you might of seen have been added for Sirius error which we will eventually replace. Hope this makes sense happy to talk through it

kate-49: sorry I was talking about thousand-character-limit here, I noticed for the rest of the validation errors you tend to camel case, e.g. StartOrEndDate/LengthOfAward, I was just wondering if that was a decision that had been made for this service?

Nicholassully: Ah sorry that is how https://github.com/go-playground/validator/blob/master/_examples/custom-validation/main.go has it and needs to be like this case

kate-49 (approved):

finance-hub/internal/server/fee_reductions.go
```diff
@@ -36,22 +39,25 @@ func (h *FeeReductionsHandler) render(v AppVars, w http.ResponseWriter, r *http.
...
-			Notes:        f.Notes,
+			Type:                     cases.Title(language.English).String(f.Type),
+			StartDate:                f.StartDate.String(),
+			EndDate:                  f.EndDate.String(),
+			DateReceived:             f.DateReceived.String(),
+			Status:                   f.Status,
+			Notes:                    f.Notes,
+			FeeReductionCancelAction: f.FeeReductionCancelAction,
+			Id:                       strconv.Itoa(f.Id),
+			ClientId:                 clientId,
```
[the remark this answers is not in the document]
Nicholassully: I need the clientId in feeReductions to make the URL correctly and in the sql query in the BE. In less I have missed something?

Nicholassully:
The PR feedback has been done in https://github.com/ministryofjustice/opg-sirius-supervision-finance-hub/pull/64

PR #60, opened 2024-05-17
pfs-93 removed two calls for api to one and unused method in shared a…

Nicholassully:
…ssignee


Nicholassully:
Sam Joseph
Not sure I understand the intention behind this change, and even if needed, the new approach is much more prone to error. You are moving from using typed functions to switch between routes to strings, which have no safety at all. If a dev mistypes newBlackendRequestz, the code won't compile. If they mistype "financeAPY" as the string, the new request method will attempt to send a request but without a url *at runtime*.

PR #65, opened 2024-05-21
PFS-29: Add manual credit

denialanderror:


test-data/seed_data.sql
```diff
@@ -2,14 +2,17 @@ SET SEARCH_PATH TO supervision_finance;
 
 INSERT INTO finance_client VALUES (1, 1, '1234', 'DEMANDED', null, 12300, 2222);
 INSERT INTO finance_client VALUES (2, 2, '1234', 'DEMANDED', null, 0, 0);
-ALTER SEQUENCE finance_client_id_seq RESTART WITH 3;
+INSERT INTO finance_client VALUES (3, 3, '1234', 'DEMANDED', null, 0, 0); -- adjust-invoice.cy.js
```
Nicholassully, line 5: Do we need the comment here?

finance-api/internal/service/ledger_entries.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	tx, err := s.tx.Begin(ctx)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err = tx.Rollback(ctx); err != nil {
+			log.Println("Error rolling back transaction:", err)
```
Nicholassully, line 27: Is it worth adding in where the error is thrown just in case we are looking through the logs in the future?

cypress/e2e/adjust-invoice.cy.js
```diff
@@ -0,0 +1,34 @@
+describe("Adjust invoice form", () => {
+    it("adds a manual credit to an invoice", () => {
+        // navigate to form
```
Nicholassully, line 3: Do we still need the comments in?

finance-hub/internal/api/adjust_invoice_test.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+func TestAdjustInvoice(t *testing.T) {
+	logger, mockClient := SetUpTest()
+	client, _ := NewApiClient(mockClient, "http://localhost:3000", "", logger)
+
+	json := `{
+            "adjustmentType": "credit write off",
+            "notes": "notes here",
+			"amount": "10000"
```
Nicholassully, line 22: Strange formatting here

Nicholassully (approved):

test-data/seed_data.sql
denialanderror: Yes. The way we have the Cypress tests set up at the moment means they don't own their own data, which means it is very easy to write a passing test in isolation that fails when run with the others as they interfere with the expected data. Adding comments in will help ensure that the seed data is only used for its intended purpose.

Ideally, the Cypress tests would create and tear down their own data but that will be a future improvement.

cypress/e2e/adjust-invoice.cy.js
denialanderror: Yes. All the assertions are being in run in one Cypress test rather than multiple, which is Cypress best practice, as starting up a test is costly. The comments make it clear what section each is testing. It's a bit of an experiment and might not be the approach we use in the future, but all these Cypress tests need reviewing anyway.

finance-hub/internal/api/adjust_invoice_test.go
denialanderror: That's weird. I think it might be Github parsing tabs over spaces because it looks formatted correctly in the IDE.

Nicholassully (approved):

PR #67, opened 2024-05-23
PFS-71 Create reject pending invoice adjustment functionality

josephsmith0705:


Makefile
```diff
@@ -54,7 +54,7 @@ start-and-seed:
 	docker compose run --rm --build finance-migration
 	docker compose exec sirius-db psql -U user -d finance -a -f ./seed_data.sql
 
-cypress: setup-directories start-and-seed
+cypress: setup-directories clean start-and-seed
```
josephsmith0705, line 57: Take down the current containers, otherwise trying to seed will cause errors as primary keys already exist

finance-api/cmd/api/update_pending_invoice_adjustment.go
```diff
@@ -1,14 +1,23 @@
...
 import (
+	"encoding/json"
 	"net/http"
 	"strconv"
 )
 
 func (s *Server) updatePendingInvoiceAdjustment(w http.ResponseWriter, r *http.Request) {
+	var body struct {
+		Status string `json:"status"`
+	}
```
denialanderror, line 12: If you move this struct into the shared package, you can then add validation to it, such as `oneof` to guard against values other than approved or rejected.

finance-hub/internal/server/submit_reject_pending_invoice_adjustment.go
denialanderror, line 1: I think this and the approve handlers can be combined and just pass through the status in the URL instead of having a whole new handler for it

finance-hub/internal/server/submit_reject_pending_invoice_adjustment.go
```diff
@@ -0,0 +1,28 @@
...
+
+type SubmitRejectPendingInvoiceAdjustmentHandler struct {
+	router
+}
+
+func (h *SubmitRejectPendingInvoiceAdjustmentHandler) render(v AppVars, w http.ResponseWriter, r *http.Request) error {
+	ctx := getContext(r)
+	var (
+		ledgerId, _ = strconv.Atoi(r.PathValue("ledgerId"))
+		status      = "rejected"
```
denialanderror, line 18: Personally, I'd prefer the status to be passed around in upper case, which generally signifies its a constant and a placeholder.

denialanderror (changes requested):

finance-hub/internal/server/route.go
```diff
@@ -111,6 +111,10 @@ func (r route) getSuccess(req *http.Request) string {
 		return "You have approved the credit"
 	case "approve-invoice-adjustment[WRITE OFF]":
 		return "You have approved the write off"
+	case "reject-invoice-adjustment[CREDIT]":
+		return "You have rejected the credit"
+	case "reject-invoice-adjustment[WRITE OFF]":
+		return "You have rejected the write off"
```
denialanderror, line 117: Debit adjustments now exist as well. Looks like I missed adding in the approve message so could you add both approve and reject for debit here please?

finance-hub/internal/api/update_pending_invoice_adjustment_test.go
```diff
@@ -49,7 +49,7 @@ func TestUpdatePendingInvoiceAdjustmentReturns500Error(t *testing.T) {
 
 	client, _ := NewApiClient(http.DefaultClient, svr.URL, svr.URL, logger)
 
-	err := client.UpdatePendingInvoiceAdjustment(getContext(nil), 1, 2)
+	err := client.UpdatePendingInvoiceAdjustment(getContext(nil), 1, 2, "approved")
```
denialanderror, line 52: Can you change all the instances in tests where we are using "approved" to "APPROVED", as that is the actual expected value now? Doesn't make a difference for the tests but it may save some confusion in the future.

finance-hub/internal/server/submit_approve_pending_invoice_adjustment.go
denialanderror, line 1: Can you delete this file now?

denialanderror (changes requested):

denialanderror (approved):

PR #68, opened 2024-05-23
Pfs 35 add write off

denialanderror:


cypress/e2e/adjust-invoice.cy.js
```diff
@@ -28,7 +28,19 @@ describe("Adjust invoice form", () => {
         cy.get('.govuk-button').click();
 
         // navigation and success message
-        cy.url().should('include', "clients/3/invoices?success=CREDIT_MEMO");
+        cy.url().should('include', "clients/3/invoices?success=invoice-adjustment[CREDIT%20MEMO]");
```
kate-49, line 31: Is there a reason you've gone for a space in this value? It looks a bit simpler when it was CREDIT_MEMO

finance-api/internal/service/ledger_entries.go
```diff
@@ -12,53 +12,54 @@ import (
...
-		return err
-	}
+func (s *Service) validateAdjustmentAmount(adjustment *shared.CreateLedgerEntryRequest, balance store.GetInvoiceBalanceRow) error {
 	switch adjustment.AdjustmentType {
 	case shared.AdjustmentTypeAddCredit:
-		if int32(adjustment.Amount)-b.Outstanding > b.Initial {
-			return shared.BadRequest{Field: "Amount", Reason: fmt.Sprintf("Amount entered must be equal to or less than £%d", (b.Initial+b.Outstanding)/100)}
+		if int32(adjustment.Amount)-balance.Outstanding > balance.Initial {
+			return shared.BadRequest{Field: "Amount", Reason: fmt.Sprintf("Amount entered must be equal to or less than £%d", (balance.Initial+balance.Outstanding)/100)}
+		}
```
kate-49, line 47: Is there any risk of us having issues with rounding as this is not money type?

kate-49 (approved):

cypress/e2e/adjust-invoice.cy.js
denialanderror: Ideally, we'd use an underscore everywhere but for some unknown reason, these values are stored in the database with spaces instead. We could have these with the underscore in but it means adding another translation function to the enum.

finance-api/internal/service/ledger_entries.go
denialanderror: It will round it to the nearest pound, which is all that's really needed in this instance.

PR #70, opened 2024-05-29
PFS-30: Add debit

denialanderror:


finance-api/internal/store/queries/ledger_allocation.sql
```diff
@@ -11,7 +11,6 @@ WITH filtered_ledger_allocation AS (
     from ledger l
              inner join ledger_allocation lc on lc.ledger_id = l.id
     where l.id = $1
-      and l.type IN ('CREDIT MEMO', 'CREDIT WRITE OFF')
```
josephsmith0705, line 14: Out of interest, why are we removing this where clause?

finance-api/internal/service/ledger_entries_test.go
```diff
@@ -160,6 +160,58 @@ func TestService_ValidateAdjustmentAmount(t *testing.T) {
...
+			err: shared.BadRequest{Field: "Amount", Reason: "Amount entered must be equal to or less than £100"},
+		},
+		{
+			name: "Add Debit - valid",
+			adjustment: &shared.CreateLedgerEntryRequest{
+				AdjustmentType: shared.AdjustmentTypeAddDebit,
+				Amount:         22000,
+			},
+			balance: store.GetInvoiceBalanceRow{
+				Initial:     16000, // invoice calculated at £160 instead of £320 in error at creation
```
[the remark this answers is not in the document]
denialanderror: It's covering a test scenario from the ticket. The comment doesn't add anything without that context though, so I'll remove it.

finance-api/internal/store/queries/ledger_allocation.sql
denialanderror: We are already addressing the ledger allocation by id so the filter isn't necessary. I expect it was copy pasted from the fetch query, where it is needed.

denialanderror: I've added this back in (and refactored the query) as it is a useful guard against the query being used incorrectly.

josephsmith0705 (approved):

PR #74, opened 2024-06-03
PFS-73 Create pending invoice adjustment task

josephsmith0705:


finance-api/internal/store/queries/ledger.sql
```diff
@@ -35,4 +35,5 @@ SELECT nextval('ledger_allocation_id_seq'),
        $3,
        'PENDING',
        $4
-FROM ledger;
\ No newline at end of file
+FROM ledger
+returning (SELECT reference FROM invoice WHERE id = invoice_id);
```
denialanderror, line 39: I'm having second thoughts about this approach now. It avoids making a second call to the database but it makes the query more confused in terms of what it is returning. I'm not sure I would expect the get the invoice reference back when I create a ledger, or at least, not _just_ the reference.

I think if we stick with this solution, the reference variable should at least be renamed to `invoiceReference`, so it is clear what it is a reference to.

finance-api/internal/service/ledger_entries.go
```diff
@@ -29,14 +29,14 @@ func (s *Service) CreateLedgerEntry(clientId int, invoiceId int, ledgerEntry *sh
...
+	invoiceReference, err := s.store.CreateLedger(ctx, params)
 
 	if err != nil {
 		log.Println("Error creating ledger entry: ", err)
-		return err
+		return nil, err
 	}
 
-	return nil
+	return &shared.InvoiceAdjustment{InvoiceRef: invoiceReference}, nil
```
denialanderror, line 39: I would return this as a string or a new struct containing a string rather than reusing the InvoiceAdjustment struct. Golang defaults omitted values, so although you are only populating the reference, all the other values will be included, just as their defaults, which would be confusing as an API.

denialanderror (changes requested):

finance-hub/internal/api/adjust_invoice.go
```diff
@@ -36,6 +36,17 @@ func (c *ApiClient) AdjustInvoice(ctx Context, clientId int, invoiceId int, adju
 	defer resp.Body.Close()
 
 	if resp.StatusCode == http.StatusCreated {
+		var response shared.InvoiceAdjustment
```
denialanderror, line 39: This is now returning a `shared.InvoiceReference`

denialanderror (approved):

PR #81, opened 2024-06-17
pfs-94 changed value of a status

Nicholassully:


finance-hub/internal/server/invoices.go
```diff
@@ -101,9 +102,26 @@ func (h *InvoicesHandler) transformLedgers(ledgers []shared.Ledger, caser cases.
...
+		return "Manual Credit"
+	case shared.AdjustmentTypeAddDebit.Key():
+		return "Manual Debit"
+	}
+
+	words := strings.Fields(transactionType)
+	for i, word := range words {
+		words[i] = strings.ToUpper(string(word[0])) + strings.ToLower(word[1:])
+	}
+	return strings.Join(words, " ")
```
denialanderror, line 126: Use `golang.org/x/text/cases` instead of doing this manually.

denialanderror (changes requested):

finance-hub/internal/server/invoices.go
Nicholassully: done

denialanderror (approved):

PR #84, opened 2024-06-19
pfs-104: Adds in public schema for assignees table access [DO NOT MERGE]

denialanderror:
These would be the proposed changes from the Payments side if we were going to query the assignees table directly. The user and roles changes would be on the infra side and the user would be passed in in environment variables as usual.

Technically, these changes would work without the role restrictions as the sole DB user has access to everything anyway.

An alternative to these changes (assuming we still go with a DB solution as opposed to an API) would be to maintain two DB connections, and read the schemas separately with two separate SQLC instances. This would potentially be slightly more complex architecturally but would remove the need for the vast majority of these changes, allow migrations to still be scoped by search path to prevent unwanted changes, etc.

A third, slightly more involved suggestion would be to move the assignees table out into a new schema - e.g. `identity` - and then use  search path to only access those two schemas. This is probably unnecessary but could add another level of security, even if set at a user level as opposed to connection.

### DIFF
--- a/docker/finance-migration/migrate.sh
+++ b/docker/finance-migration/migrate.sh
@@ -3,7 +3,7 @@
 set -e
 set -o pipefail
 
-DB_CONN_STR="postgres://$DB_USER:$DB_PASSWORD@$DB_CONNECTION/$DB_NAME?search_path=$DB_SCHEMA"
+DB_CONN_STR="postgres://$DB_USER:$DB_PASSWORD@$DB_CONNECTION/$DB_NAME"
 COMMAND=$1
 
 check_var() {
@@ -19,7 +19,6 @@ check_var DB_USER $DB_USER
 check_var DB_PASSWORD $DB_PASSWORD
 check_var DB_CONNECTION $DB_CONNECTION
 check_var DB_NAME $DB_NAME
-check_var DB_SCHEMA $DB_SCHEMA
 check_var COMMAND $1
 
 /usr/local/bin/goose -dir /database postgres $DB_CONN_STR $COMMAND
--- a/finance-api/internal/service/add_fee_reduction.go
+++ b/finance-api/internal/service/add_fee_reduction.go
@@ -50,13 +50,13 @@ func (s *Service) AddFeeReduction(id int, data shared.AddFeeReduction) error {
 
 	transaction := s.store.WithTx(tx)
 
-	var feeReduction store.FeeReduction
+	var feeReduction store.SupervisionFinanceFeeReduction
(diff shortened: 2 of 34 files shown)


docker/finance-migration/migrate.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 set -o pipefail
 
-DB_CONN_STR="postgres://$DB_USER:$DB_PASSWORD@$DB_CONNECTION/$DB_NAME?search_path=$DB_SCHEMA"
+DB_CONN_STR="postgres://$DB_USER:$DB_PASSWORD@$DB_CONNECTION/$DB_NAME"
```
denialanderror, line 6: We can't address the DB via search path in a multi-schema scenario because of how SQLC parses migration files to produce its structs. This will work fine but it does mean there would be no protection against Payments modifying the public (or any other) schema.

finance-api/internal/service/add_fee_reduction.go
```diff
@@ -50,13 +50,13 @@ func (s *Service) AddFeeReduction(id int, data shared.AddFeeReduction) error {
 
 	transaction := s.store.WithTx(tx)
 
-	var feeReduction store.FeeReduction
+	var feeReduction store.SupervisionFinanceFeeReduction
```
denialanderror, line 53: Due to requiring fully qualified paths (i.e. including the schema) in the schema files for SQLC, the generated models are created with the full path as name. This is a bit unwieldy but not really an issue, and we tend not to use the generated models anyway.

finance-api/internal/service/add_fee_reduction_test.go
```diff
@@ -36,6 +36,7 @@ func (suite *IntegrationSuite) TestService_AddFeeReduction() {
 	conn := suite.testDB.GetConn()
 
 	conn.SeedData(
+		"SET SEARCH_PATH to supervision_finance;",
```
denialanderror, line 39: Could potentially roll this into the `SeedData` function.

finance-api/main.go
```diff
@@ -76,7 +76,7 @@ func main() {
 	// Open a connection to the PostgreSQL database
 	ctx := context.Background()
 
-	conn, err := pgx.Connect(ctx, fmt.Sprintf("postgresql://%s:%s@%s/%s?search_path=supervision_finance", dbUser, dbPassword, dbConn, pgDb))
+	conn, err := pgx.Connect(ctx, fmt.Sprintf("postgresql://%s:%s@%s/%s", dbUser, dbPassword, dbConn, pgDb))
```
denialanderror, line 79: The intention here would be to connect with a "payments user" and access would be restricted by role, rather than search path.

PR #87, opened 2024-06-21
pfs-77 frontend ui

Nicholassully:


cypress/e2e/billing-history.cy.js
```diff
@@ -0,0 +1,14 @@
+describe("Billing History Tab", () => {
+    it("the feed of the billing history show", () => {
+        cy.visit("/clients/1/invoices");
+        cy.get('[data-cy="billing-history"]').click()
```
denialanderror, line 4: Interact with the page using user-facing attributes where possible. The tab has text, so find the nav element with the text "Billing History" and click that instead.

cypress/e2e/billing-history.cy.js
```diff
@@ -0,0 +1,14 @@
...
+        cy.get('.moj-timeline').first().contains("Write off applied to");
+        cy.get('.govuk-link').first().click();
+        cy.url().should('include', "clients/1/invoices");
+    });
+
+    it("no history shows correct message", () => {
+        cy.visit("/clients/2/billing-history");
+        cy.contains('[data-cy="no-billing-history"]', "No billing history for this client");
```
denialanderror, line 12: Is the `data-cy` tag here necessary? Just checking the text is present would be enough.

denialanderror (changes requested):

denialanderror (approved):

PR #92, opened 2024-06-27
pfs-82 added pending adjustment

Nicholassully:


finance-api/internal/store/queries/ledger.sql
```diff
@@ -14,7 +14,7 @@ WHERE l.id = $2;
 -- name: CreateLedger :one
 WITH fc AS (SELECT id FROM finance_client WHERE client_id = $1),
      ledger AS (
-         INSERT INTO ledger (id, datetime, amount, notes, type, finance_client_id, reference, method, status)
+         INSERT INTO ledger (id, datetime, amount, notes, type, finance_client_id, reference, method, status, createddate)
```
denialanderror, line 17: ledger already has a created date, it's just (unhelpfully) called datetime. It would be better to add in a migration to rename than add another field, but we can't do that without modifying Sirius at the stage.

Nicholassully: I tried using the created date but was seeing that the order of the timeline was inconsistent and it did not have the time part on the field. I used the datetime field that was already on the table to try to make the order correct.

Nicholassully: Never mind I understand what you meant now and have removed it

denialanderror (approved):

PR #95, opened 2024-07-02
pfs-58 sqlc vet into git actions

Nicholassully:


finance-api/internal/store/sqlc.yaml
```diff
@@ -7,4 +7,11 @@ sql:
       go:
         package: "store"
         sql_package: "pgx/v5"
-        out: "."
\ No newline at end of file
+        out: "."
+    rules:
+      - check-num-of-params
+rules:
+  - name: check-num-of-params
```
denialanderror, line 14: `check` is redundant. Also, name the rule to it is clear what has broken it. `num-of-params` doesn't tell me if I have too many or too few.

finance-api/internal/store/sqlc.yaml
```diff
@@ -7,4 +7,11 @@ sql:
...
+    message: "do you need all the parameters"
```
denialanderror, line 15: Failing the vetting fails the build, so the message has to be a failure statement, not a question. If the answer to "do you need all the parameters" was "yes", the build is still going to fail!

finance-api/internal/store/sqlc.yaml
```diff
@@ -7,4 +7,11 @@ sql:
...
+    rule: |
+      query.params.size() > 7
```
denialanderror, line 17: Is there a reason why you've gone for 7?

denialanderror (changes requested):
There's not too much value adding vetting in if there is only one rule. See if there are other Postgres best practices you can enforce.

Also, the one rule that would add the most value is not being used: [sqlc-db-prepare](https://docs.sqlc.dev/en/latest/howto/vet.html#sqlc-db-prepare)

finance-api/internal/store/sqlc.yaml
Nicholassully: We have an insert that needs 6 params and I wanted to add in an example of a rule for when we decide what we need.

finance-api/internal/store/sqlc.yaml
```diff
@@ -3,8 +3,21 @@ sql:
...
\ No newline at end of file
+        out: "."
+    rules:
+      - no-more-than-seven-params
+      - postgresql-query-too-costly
+      - sqlc/db-prepare
+rules:
+  - name: no-more-than-seven-params
+    message: "too many parameters"
+    rule: query.params.size() > 14
```
denialanderror, line 20: It says "seven" but the rule is 14

finance-api/internal/store/sqlc.yaml
```diff
@@ -3,8 +3,21 @@ sql:
...
+  - name: postgresql-query-too-costly
+    message: "Too costly"
+    rule: postgresql.explain.plan.total_cost > 50.0
```
denialanderror, line 23: Is there a reasoning for choosing 50?

denialanderror (changes requested):

denialanderror (approved):

PR #98, opened 2024-07-04
pfs-66 changing the sirius route

Nicholassully:


denialanderror (dismissed):
Approved, but a cleaner approach would have been to put the prefix in the `newSiriusRequest` function, rather than requiring it to be passed in each time the function is called.

denialanderror (approved):

PR #99, opened 2024-07-04
pfs-99: Tests refactored and axe added to each page. Form a11y needs …

denialanderror:
…fixing.


cypress/e2e/add-fee-reduction.cy.js
```diff
@@ -1,29 +1,30 @@
...
+        cy.contains('a', "Award a fee reduction").click();
+
+        // ensure validation is configured correctly
+        cy.contains('.govuk-button', "Save and continue").click();
+        cy.get('.govuk-error-summary').contains("Enter a reason for awarding fee reduction");
+        cy.get(".govuk-form-group--error").should('have.length', 5);
+
+        // successfully submit
+        cy.get('#f-FeeType').contains(".govuk-radios__item", "Hardship").click();
+        cy.get('#f-StartYear').find('input').type("2000c");
```
kate-49, line 14: This c seems weird?

cypress/e2e/finance_hub.cy.js
```diff
@@ -26,6 +26,10 @@ describe("Finance Hub", () => {
             cy.get('[data-cy="invoices"]').click();
             cy.url().should("contain", "invoices");
             cy.contains(".govuk-heading-l", "Invoices");
+
+            cy.get('[data-cy="billing-history"]').click();
```
kate-49, line 30: Is there a reason you chose this attribute over an id?

kate-49 (dismissed):

cypress/e2e/finance_hub.cy.js
denialanderror: It was mainly filling in a missing test by following the existing pattern in the file.

Cypress best practice states that you should choose attributes over ids as they are isolated from changes and HTML semantics. Personally I disagree and think Cypress have things backwards - tests should be asserting on what is on the page (e.g. text), otherwise you are just testing you've set your attributes, not that it is working for the user.

Nicholassully (approved):

PR #102, opened 2024-07-18
PFS-119 Fee reduction credit applied event

josephsmith0705:
Added fee reduction credit applied event to billing history

### DIFF
--- a/finance-api/internal/service/billing_history.go
+++ b/finance-api/internal/service/billing_history.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"github.com/opg-sirius-finance-hub/finance-api/internal/store"
 	"github.com/opg-sirius-finance-hub/shared"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"math"
 	"sort"
 	"strconv"
@@ -37,9 +39,19 @@ func (s *Service) GetBillingHistory(ctx context.Context, clientId int) ([]shared
 	if err != nil {
 		return nil, err
 	}
-	history := processAllocations(allocationsByLedger, int(startingBalance.Outstanding))
-	sortHistoryByDate(history)
+	history := processPendingAllocations(allocationsByLedger, int(startingBalance.Outstanding))
 
+	approvedAllocations, err := s.store.GetApprovedLedgerAllocations(ctx, int32(clientId))
+	if err != nil {
+		return nil, err
+	}
+
+	allocationsByLedger = aggregateApprovedAllocations(approvedAllocations, strconv.Itoa(clientId))
+	approvedAllocationsHistory := processApprovedAllocations(allocationsByLedger, int(startingBalance.Outstanding))
+
+	history = append(history, approvedAllocationsHistory...)
+
+	sortHistoryByDate(history)
 	return computeBillingHistory(history), nil
 }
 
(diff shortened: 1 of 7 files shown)


finance-api/internal/service/billing_history.go
```diff
@@ -165,18 +170,48 @@ func processFeeReductionEvents(feEvents []store.GetFeeReductionEventsRow) []hist
 					},
 				},
 			}
+			history = append(history, historyHolder{
+				billingHistory:    bh,
+				balanceAdjustment: 0,
+			})
+		}
+		if fe.Status.String == "APPROVED" {
```
denialanderror, line 178: This works for now but would end up complicating things when we need to do the billing history for payment events, as we would then have to exclude allocations created as a result of fee reductions to prevent them from being double counted.

denialanderror (changes requested):

migrations/1_baseline.sql
```diff
@@ -146,7 +146,7 @@ create table invoice
     enddate           date        not null,
     amount            integer     not null,
     supervisionlevel  varchar(255) default NULL::character varying,
-    confirmeddate     date,
+    confirmeddate     timestamp(0),
```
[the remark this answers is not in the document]
josephsmith0705: I've changed a few of the dates to timestamps because their order in the billing history was getting messed up. If I'm remembering right, we seemed to have a workaround where order by date, and if the dates are the same, the event that had a lower balance was assumed to be later. However, this didn't work for the fee reduction applied events, as they increased the balance, so were always moved earlier in the order than they were done. I've moved all these changes from date to timestamp to their own migrations, so it's clearer what's going on

finance-api/internal/service/billing_history.go
josephsmith0705: As right now this only fetches allocations that are linked to a fee reduction, I'd imagine we could just have a separate query to fetch non-fee reduction related allocations, and then process them separately? It'd be less performant but keep the code simple.

migrations/20240820180543_fee_reduction_applied_dates_to_timestamps.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+ALTER TABLE invoice ALTER COLUMN createddate TYPE timestamp(0);
+ALTER TABLE invoice ALTER COLUMN confirmeddate TYPE timestamp(0);
```
denialanderror, line 3: We have to be cautious about changing columns that exist in Sirius with migrations on our end, as that may cause inadvertent issues with Sirius, either with mismatched types or conflicts with future migrations.

I think in this case, there shouldn't be a type issue as the type in the Invoice entity in Sirius is DateTimeImmutable anyway, but check this a) works in adhoc (if you haven't already) and b) that applying this doesn't affect creating migrations in Doctrine. I'm not sure whether Doctrine looks at the state of the database or the entities when generating migrations.

josephsmith0705: I've tested in adhoc and found that, for this to work with Sirius the invoice table's confirmed date needs to be updated from date_immutable to datetime_immutable (which I've done in [this pr](https://github.com/ministryofjustice/opg-sirius/pull/10478). As far as I can tell, there shouldn't be an issue with merging this now ahead of time, I've tested this PR on adhoc also and it doesn't seem to affect the current functionality

josephsmith0705: After further discussion we don't need invoice confirmeddate, so the only columns being changed here are added by the finance hub

migrations/20240410094117_ledger_created_fields.sql
```diff
@@ -1,7 +1,7 @@
 -- +goose Up
-ALTER TABLE ledger ADD COLUMN createddate date;
-ALTER TABLE ledger ADD COLUMN createdby_id int;
+ALTER TABLE ledger ADD COLUMN created_at date;
```
denialanderror, line 2: Presumably you want this one to be timestamp too?

denialanderror (approved):

PR #109, opened 2024-07-30
PFS-125 add manual invoice validation fixes

josephsmith0705:


shared/nillable_date.go
denialanderror, line 1: Instead of creating one of these for everything you want to be nillable, you can use generics instead.

```
type Nillable[T any] struct {
	Value T
	Valid bool
}
```

And then you create one like `Nillable[int]{Value: 1, Valid: true}` or `Nillable[shared.Date]{Value: shared.NewDate("01/02/2003"), Valid: true}`.

shared/add_manual_invoice.go
denialanderror, line 1: Are `nillable-date-required` and `nillable-int-required` still needed? For the amount field for instance, the greater than or less than checks already pass through nil values, so the required check isn't doing anything.

finance-api/internal/service/add_manual_invoice.go
```diff
@@ -10,21 +10,52 @@ import (
...
 )
 
+func processInvoiceData(data shared.AddManualInvoice) shared.AddManualInvoice {
+	switch data.InvoiceType {
+	case shared.InvoiceTypeAD:
+		data.Amount = shared.Nillable[int]{Value: 100, Valid: true}
+		data.StartDate = data.RaisedDate
+		data.EndDate = data.RaisedDate
+	case shared.InvoiceTypeS2, shared.InvoiceTypeB2:
+		if data.RaisedYear.Value != 0 {
```
denialanderror, line 20: This isn't your implementation but I don't think it makes sense to send over the raised year rather than just the raised date. Putting the raised year in the form for some of the invoice types is an UI implementation, not a concern of the API.

shared/add_manual_invoice.go
josephsmith0705: The date required one is still needed, but I've removed the int required validation

denialanderror (approved):

PR #132, opened 2024-09-02
PFS-37 write off reversal invoice adjustment

josephsmith0705:


finance-api/internal/service/invoices.go
```diff
@@ -97,13 +99,16 @@ func (ib *invoiceBuilder) addLedgerAllocations(ilas []store.GetLedgerAllocations
...
-			} else if il.Type == "CREDIT WRITE OFF" {
+			} else if il.Type == "CREDIT WRITE OFF" && !writeOffReversed {
 				metadata.contextType = "Write-off"
 			}
 		}
-		if metadata.contextType == "" && il.Type == "CREDIT WRITE OFF" {
+		if metadata.contextType == "" && il.Type == "CREDIT WRITE OFF" && !writeOffReversed {
 			metadata.contextType = "Write-off pending"
 		}
+		if il.Type == "WRITE OFF REVERSAL" {
```
Nicholassully, line 109: Is there a test for this, also how come we are not checking this before the other `if` conditions as we change the value after

cypress/e2e/customer-credit-balance.cy.js
```diff
@@ -29,7 +29,7 @@ describe("Customer credit balance", () => {
         cy.contains("AD77777/24").click();
         cy.get("table#ledger-allocations > tbody > tr").should("have.length", 3);
 
-        const amounts = ["£30", "£-30", "£100"];
+        const amounts = ["£30", "£30", "£100"];
```
denialanderror, line 32: Why has the minus been removed from the unapplied amount?

finance-api/internal/service/add_invoice_adjustment.go
```diff
@@ -59,18 +65,30 @@ func (s *Service) validateAdjustmentAmount(adjustment *shared.AddInvoiceAdjustme
...
-func (s *Service) calculateAdjustmentAmount(adjustment *shared.AddInvoiceAdjustmentRequest, balance store.GetInvoiceBalanceDetailsRow) int32 {
+func (s *Service) calculateAdjustmentAmount(adjustment *shared.AddInvoiceAdjustmentRequest, balance store.GetInvoiceBalanceDetailsRow, customerCreditBalance int32) int32 {
 	switch adjustment.AdjustmentType {
 	case shared.AdjustmentTypeWriteOff:
 		return balance.Outstanding
 	case shared.AdjustmentTypeDebitMemo:
 		return -int32(adjustment.Amount)
+	case shared.AdjustmentTypeWriteOffReversal:
+		if balance.WrittenOff {
+			if balance.Initial > customerCreditBalance {
```
denialanderror, line 86: Not sure this is correct. A write off is calculated as the outstanding balance, not the initial amount. e.g. an invoice of £100 that has a partial payment of £20 would have a write off of £80.

finance-api/internal/service/invoice_adjustments.go
```diff
@@ -18,7 +19,7 @@ func (s *Service) GetInvoiceAdjustments(ctx context.Context, clientId int) (*sha
 			Id:             int(ia.ID),
 			InvoiceRef:     ia.InvoiceRef,
 			RaisedDate:     shared.Date{Time: ia.RaisedDate.Time},
-			Amount:         int(ia.Amount),
+			Amount:         int(math.Abs(float64(ia.Amount))),
```
denialanderror, line 22: Same question as above re: this no longer being negative.

If this change is a requirement, this wouldn't be the right place to call abs. The API should return the data in as true a representation as possible. If the UI wants to display all values as positive, it should do that transformation itself.

denialanderror (changes requested):

finance-api/internal/service/invoice_adjustments.go
josephsmith0705: This was a request from Alex, that the transaction type should indicate to the users whether it'll be positive or negative, so the minus is not needed - I'll move it to the frontend, rather than the api

finance-api/internal/service/invoices.go
josephsmith0705: Good spot, I've added a test and moved it up above the other conditions

finance-api/internal/service/invoice_adjustments.go
denialanderror: Oh okay. That's a bit weird, seeing as we have tickets in the backlog for moving the minus from "£-" to "-£".

finance-api/internal/service/add_invoice_adjustment.go
```diff
@@ -16,6 +16,16 @@ func (s *Service) AddInvoiceAdjustment(ctx context.Context, clientId int, invoic
 		return nil, err
 	}
 
+	clientInfo, err := s.store.GetAccountInformation(ctx, int32(clientId))
+	if err != nil {
+		return nil, err
+	}
+
+	writeOffAmount, err := s.store.GetMostRecentApprovedWriteOffAmount(ctx, int32(invoiceId))
```
denialanderror, line 24: I'd be tempted to include the write off amount in `GetInvoiceBalanceDetails` instead of another query. It already fetches a boolean for if it the invoice has been written off so you could just change that to an int and use zero of the false value.

finance-hub/internal/util/translate.go
```diff
@@ -15,8 +15,9 @@ var validationMappings = map[string]map[string]pair{
 		"valid-enum": pair{"AdjustmentType", "Select the adjustment type"},
 	},
 	"AdjustmentNotes": {
-		"required":            pair{"AdjustmentNotes", "Enter a reason for adjustment"},
-		"stringLengthTooLong": pair{"AdjustmentNotes", "Reason for manual credit must be 1000 characters or less"},
+		"required":                 pair{"AdjustmentNotes", "Enter a reason for adjustment"},
+		"stringLengthTooLong":      pair{"AdjustmentNotes", "Reason for manual credit must be 1000 characters or less"},
+		"thousand-character-limit": pair{"Notes", "Reason for adjustment must be 1000 characters or less"},
```
denialanderror, line 20: There appears to be two validation messages for the same error

denialanderror (approved):

PR #149, opened 2024-09-30
PFS-130 Process moto card payments upload

josephsmith0705:


finance-api/internal/awsclient/client.go
```diff
@@ -0,0 +1,43 @@
+package awsclient
```
denialanderror, line 1: `awsclient` implies this is the client for all AWS integration, but we use a different client for EventBridge. I'd suggest renaming it. `filestorage`? `storage`?

finance-api/internal/service/finance_clients.go
```diff
@@ -17,3 +19,16 @@ func (s *Service) GetAccountInformation(ctx context.Context, id int) (*shared.Ac
...
 	}, nil
 }
+
+func (s *Service) UpdateClient(ctx context.Context, id int, caseRecNumber string) error {
+	err := s.store.UpdateClient(ctx, store.UpdateClientParams{
+		Caserecnumber: pgtype.Text{String: caseRecNumber, Valid: true},
+		ClientID:      int32(id),
+	})
+
+	if err != nil {
```
denialanderror, line 29: Instead of this conditional, you can just `return err`, as if it is nil, you return it later anyway.

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/opg-sirius-finance-hub/finance-api/internal/awsclient"
+	"github.com/opg-sirius-finance-hub/finance-api/internal/store"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func (s *Service) ProcessFinanceAdminUpload(ctx context.Context, bucketName string, key string) error {
+	client, _ := awsclient.NewClient(ctx)
```
denialanderror, line 17: Clients are typically designed to be created once at start up and then reused, rather than initialised on each function call. You can add it to the Service struct in the same way I did with the EventBridge client.

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -0,0 +1,134 @@
...
+
+	output, err := client.GetObject(ctx, &s3.GetObjectInput{
```
denialanderror, line 19: It would be nice to avoid leaking AWS implementation into the service, e.g. if the call on the client just took a key and bucket name as strings and handled the types and parameter parsing internally.

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -0,0 +1,134 @@
...
+		amount = amount + "0"
+	} else if index == -1 {
+		amount = amount + "00"
+	}
+
+	intAmount, _ := strconv.Atoi(strings.Replace(amount, ".", "", 1))
+	return intAmount
+}
+
+func (s *Service) processMotoCardPaymentsUploadLine(ctx context.Context, record []string) error {
```
denialanderror, line 58: This will all need wrapping in a transaction. Otherwise, if it failed halfway through, it could create a ledger in the DB but not associate it with any payments.

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -0,0 +1,134 @@
...
+		return err
+	}
+
+	invoices, err := s.store.GetInvoicesForCaseRecNumber(ctx, pgtype.Text{String: courtReference, Valid: true})
+	if err != nil {
+		return err
+	}
+
+	for _, invoice := range invoices {
+		if !(invoice.Amount == invoice.Received) && amount > 0 {
```
denialanderror, line 101: If you filter out the closed/paid invoices in the query (e.g. with a HAVING clause), you won't need this check

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if ledgerId != 0 {
+		return nil
+	}
+
+	ledgerId, err = s.store.CreateLedgerForCaseRecNumber(ctx, store.CreateLedgerForCaseRecNumberParams{
+		Caserecnumber: pgtype.Text{String: courtReference, Valid: true},
+		Amount:        int32(amount),
+		Type:          "MOTO card payment",
+		Status:        "APPROVED",
```
denialanderror, line 86: We will be getting rid of ledger status eventually but I think the correct status her to keep in line with existing payments is "CONFIRMED"

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -0,0 +1,134 @@
...
+			err := s.processMotoCardPaymentsUploadLine(ctx, record)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func parseAmount(amount string) int {
```
denialanderror, line 45: If you return int32 here, you won't have to do so much type conversion in the process function

finance-api/internal/store/queries/invoices.sql
```diff
@@ -16,6 +16,22 @@ WHERE fc.client_id = $1
 GROUP BY i.id, i.raiseddate
 ORDER BY i.raiseddate DESC;
 
+-- name: GetInvoicesForCaseRecNumber :many
+SELECT i.id,
+       i.raiseddate,
+       i.reference,
+       i.amount,
+       COALESCE(SUM(la.amount), 0)::INT    received,
+       COALESCE(MAX(fr.type), '')::VARCHAR fee_reduction_type
```
denialanderror, line 25: Is fee reduction used in processing this file?

migrations/20241002141959_finance_client_caserecnumber.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+ALTER TABLE finance_client ADD COLUMN caserecnumber varchar(255);
```
denialanderror, line 2: You may want to add an index along with this if that's how we are querying by

denialanderror (changes requested):
I'd rename references to CaseRecNumber number to "CourtRef". Casrec was the old system in place before Sirius, so it would be good to move away from that where possible.

finance-api/internal/store/queries/invoices.sql
josephsmith0705: You're right, it's not - there were a bunch of unused returns here that I've reduced to only what is needed

denialanderror (approved):
Re: transactions, we should probably do some performance testing at some point. This approach means that if one line fails, the previous successful ones aren't rolled back, which is good, but it also means we are doing a database hit for each line. We could instead batch the changes and commit e.g. every 50 rows if performance is being impacted.

PR #152, opened 2024-10-09
PFS-126: Allocation timeline

denialanderror:
Refactors billing history to group transaction events, allowing us to better present ledgers with multiple allocations and unapplies.

### DIFF
--- a/finance-api/internal/service/billing_history.go
+++ b/finance-api/internal/service/billing_history.go
@@ -153,13 +153,15 @@ func processFeeReductionEvents(feEvents []store.GetFeeReductionEventsRow) []hist
 // processLedgerAllocations takes an array of allocations and groups them by ledger, which defines a single billing event.
 // A ledger is always for a single transaction type but may have multiple allocations associated with it.
 func processLedgerAllocations(allocations []store.GetLedgerAllocationsForClientRow, clientID int) []historyHolder {
-	historyByLedger := make(map[int32]historyHolder)
+	historyByLedger := make(map[int32]*historyHolder)
 
 	for _, allocation := range allocations {
 		var (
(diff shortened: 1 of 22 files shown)


finance-api/internal/service/billing_history.go
```diff
@@ -118,84 +128,149 @@ func processFeeReductionEvents(feEvents []store.GetFeeReductionEventsRow) []hist
...
 		if history[i].billingHistory.Date.Time.Equal(history[j].billingHistory.Date.Time) {
-			if _, ok := history[i].billingHistory.Event.(shared.FeeReductionApplied); ok {
-				return history[i].billingHistory.OutstandingBalance < history[j].billingHistory.OutstandingBalance
+			// reapplies should apply after if they are the result of a transaction event
+			if _, ok := history[i].billingHistory.Event.(shared.TransactionEvent); ok {
+				return history[j].billingHistory.Event.GetType() == shared.EventTypeReappliedCredit
 			}
-			return history[i].billingHistory.OutstandingBalance > history[j].billingHistory.OutstandingBalance
+			// transaction events and reapplies should apply after the event that causes them
+			return history[i].billingHistory.Event.GetType() != shared.EventTypeReappliedCredit
```
josephsmith0705, line 254: Isn't this doing the same as the above, line 251?

denialanderror: Not quite. If `i` is a transaction event but `j` is a reapply, we don't swap, as the reapply event would have been caused by the transaction event (e.g. a fee reduction applied). Otherwise, we swap if `i` isn't a reapply, as non-transaction events should occur before transaction events.

josephsmith0705 (approved):

PR #154, opened 2024-10-21
pfs-135: Hooks billing history up with the user cache

denialanderror:


finance-hub/internal/api/get_user.go
```diff
@@ -43,7 +42,7 @@ func (c *ApiClient) GetUser(ctx Context, userId int) (shared.Assignee, error) {
 	c.caches.updateUsers(users)
 	user, ok = c.caches.getUser(userId)
 	if !ok {
-		return shared.Assignee{}, errors.New("user not found")
+		return *c.caches.getAndSetPlaceholder(userId), nil
```
josephsmith0705, line 45: Am I right in thinking this means if a user can't be found for the billing history event, we set it as that placeholder, instead of throwing an error? Is that what we want / expect?

denialanderror: Yes. If an assignee id doesn't match a user in the cache, we don't know what user it relates to, so we return the placeholder, which is "Unknown User". However, we don't want to have to do that every cache call, as a missed cache hit fetches all the users again and refreshes the cache, so we instead set the placeholder to the missed id, so it will just return that next time it is fetched.

This should never be the case going forwards, but it is a useful failsafe and necessary right now, where we are using an ID of 1 for all creation events until we play the ticket to pull through user information, and there isn't an assignee with an ID of 1 in the DB.

Nicholassully (approved):

PR #184, opened 2024-12-19
pfs-187 adding in frontend dd only #minor

Nicholassully:


finance-hub/internal/api/create_direct_debit_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+func TestEmptyAccountHolderReturnsNil(t *testing.T) {
+	svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
+	defer svr.Close()
+
+	client, _ := NewApiClient(http.DefaultClient, svr.URL, svr.URL)
+
+	err := client.SubmitDirectDebit("CLIENT", "testing", "123456", "12345678")
+	assert.Nil(t, err)
+}
```
kate-49, line 98: Is there a happy path test for this file?

finance-hub/internal/api/create_direct_debit_test.go
```diff
@@ -0,0 +1,98 @@
...
+	defer svr.Close()
+
+	client, _ := NewApiClient(http.DefaultClient, svr.URL, svr.URL)
+
+	err := client.SubmitDirectDebit("CLIENT", "account Name", "12345", "12345678")
+	expectedError := apierror.ValidationError{Errors: apierror.ValidationErrors{"SortCode": map[string]string{"eqSix": ""}}}
+	assert.Equal(t, expectedError, err.(apierror.ValidationError))
+}
+
+func TestSortCodeSixCharactersAllZerosReturnsValidationError(t *testing.T) {
```
kate-49, line 68: Pretty confident Nick would say this could be a table test 😉

finance-hub/internal/api/create_direct_debit_test.go
Nicholassully: Aye this is the happy path one for the time being

finance-hub/internal/api/create_direct_debit.go
```diff
@@ -0,0 +1,40 @@
...
+		errors["AccountHolder"] = map[string]string{"required": ""}
+	}
+
+	if accountName == "" {
+		errors["AccountName"] = map[string]string{"required": ""}
+	}
+
+	if len(accountName) > 18 {
+		errors["AccountName"] = map[string]string{"gteEighteen": ""}
+	}
```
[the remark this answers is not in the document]
Nicholassully: No this value will not change its set by allPay

finance-hub/internal/api/create_direct_debit_test.go
Nicholassully: It could I will move them to a table but I think it will be harder to read lets see

finance-hub/internal/server/route.go
```diff
@@ -155,6 +159,7 @@ func IsHxRequest(req *http.Request) bool {
 
 func (r route) transformFinanceClient(person shared.Person, accountInfo shared.AccountInformation) FinanceClient {
 	return FinanceClient{
+		ClientId:           strconv.Itoa(person.ID),
```
denialanderror, line 162: This isn't going to work as part of an hx-request. If you are navigating to the page by pressing the button, this function will never be called because it is only ever called on a full page load. All other handlers that need the client ID get it from the path and add it into their vars that way.

denialanderror (changes requested):

denialanderror (approved):

PR #197, opened 2025-01-14
PFS-200 Download invoice adjustments report

josephsmith0705:


finance-api/internal/db/invoice_adjustments.go
```diff
@@ -0,0 +1,96 @@
...
+   p.caserecnumber                                                   AS "Customer number",
+   fc.sop_number                                                     AS "SOP number",
+   '="0470"'                                                            AS "Entity",
+   cc.code                                                           AS "Revenue cost centre",
+   cc.cost_centre_description                                        AS "Revenue cost centre description",
+   a.code                                                            AS "Revenue account code",
+   a.account_code_description                                        AS "Revenue account code description",
+   CONCAT(tt.fee_type, i.reference)                                  AS "Txn number",
+   tt.description                                                    AS "Txn description",
+   CASE WHEN fr.startdate IS NOT NULL AND fr.enddate IS NOT NULL THEN CONCAT(EXTRACT(YEAR FROM AGE(fr.enddate, fr.startdate)), ' year') END AS "Remission/Exemption award term",
```
denialanderror, line 25: startdate and enddate are both mandatory fields so you don't need to check both

finance-api/internal/db/invoice_adjustments.go
```diff
@@ -0,0 +1,96 @@
...
+   CASE
+       WHEN la.datetime >= DATE_TRUNC('year', la.datetime) + INTERVAL '3 months'
+           THEN CONCAT(TO_CHAR(la.datetime, 'YY'), '/', TO_CHAR((la.datetime + INTERVAL '1 YEAR'), 'YY'))
+       ELSE CONCAT(TO_CHAR((la.datetime - INTERVAL '1 YEAR'), 'YY'), '/', TO_CHAR(la.datetime, 'YY'))
+       END                                                           AS "Financial Year",
+   la.datetime                                                       AS "Approved date",
```
denialanderror, line 31: The date format we are using in other reports is `TO_CHAR(ola.datetime, 'YYYY-MM-DD')`

finance-api/internal/db/invoice_adjustments_test.go
```diff
@@ -0,0 +1,79 @@
...
+	assert.Equal(suite.T(), "12345678", results[0]["Customer number"], "Customer number - client 1")
+	assert.Equal(suite.T(), "1234", results[0]["SOP number"], "SOP number - client 1")
+	assert.Equal(suite.T(), "=\"0470\"", results[0]["Entity"], "Entity - client 1")
+	assert.Equal(suite.T(), "10482009", results[0]["Revenue cost centre"], "Cost centre - client 1")
+	assert.Equal(suite.T(), "Supervision Investigations", results[0]["Revenue cost centre description"], "Cost centre description - client 1")
+	assert.Equal(suite.T(), "4481102114", results[0]["Revenue account code"], "Account code - client 1")
+	assert.Equal(suite.T(), "INC - RECEIPT OF FEES AND CHARGES - Rem Appoint Deputy", results[0]["Revenue account descriptions"], "Account code description - client 1")
+	assert.Equal(suite.T(), fmt.Sprintf("MCR%s", client1Invoice2Ref), results[0]["Txn number and type"], "Txn number - client 1")
+	assert.Equal(suite.T(), "Manual Credit", results[0]["Txn description"], "Txn description - client 1")
+	assert.Equal(suite.T(), "<nil>", results[0]["Remission/exemption term"], "Remission/Exemption award term - client 1")
```
denialanderror, line 57: Do we want the text to be "<nil>" or do we want ""?

denialanderror (changes requested):

finance-api/internal/db/invoice_adjustments.go
josephsmith0705: They're mandatory fields when creating a remission / exemption, but this report also fetches invoice adjustments - so for these, fr.startdate will be null. But you're right, I only need to check one as the other will surely be nil if so

denialanderror (approved):

PR #204, opened 2025-01-20
PFS-239 Update validation messages

josephsmith0705:


finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -5,48 +5,65 @@ import (
...
+		return nil
 	}
 
 	failedLines, err := s.processPayments(ctx, records, detail.UploadType, detail.UploadDate)
 
 	if err != nil {
-		uploadProcessedEvent.Error = err.Error()
-		return s.dispatch.FinanceAdminUploadProcessed(ctx, uploadProcessedEvent)
+		payload := createUploadNotifyPayload(detail, err.Error(), map[int]string{})
+		err := s.notify.Send(ctx, payload)
```
denialanderror, line 45: instead of the nil check on err, you can just do `return s.notify.Send(ctx, payload)`. Same with the other instances below.

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -94,7 +111,7 @@ func getPaymentDetails(record []string, uploadType string, uploadDate shared.Dat
 
 		bankDate, err = time.Parse("2006-01-02 15:04:05", record[1])
 		if err != nil {
-			(*failedLines)[index] = "DATE_PARSE_ERROR"
+			(*failedLines)[index] = "DATE_TIME_PARSE_ERROR"
```
denialanderror, line 114: Is DATE_PARSE_ERROR still used? If not, it needs removing.

josephsmith0705: Date parse error is still used below yeah - these two different types inform the user of the required format of the date, dd/mm/yyyy or yyyy-mm-dd 00:00:00

denialanderror (approved):

PR #216, opened 2025-02-03
PFS-161 receipt & non-receipt transactions journals

josephsmith0705:


finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,158 @@
...
+                ) sl ON TRUE
+                LEFT JOIN LATERAL (
+                        SELECT CASE WHEN i.feetype IN ('GA', 'GS', 'GT') THEN '10486000' ELSE '10482009' END AS cost_centre LIMIT 1
+    ) cc ON TRUE
+        JOIN supervision_finance.transaction_type tt
+                  ON l.type = tt.ledger_type AND sl.supervision_level = tt.supervision_level
+    WHERE tt.is_receipt = false AND TO_CHAR(l.created_at, 'YYYY-MM-DD') = $1
+    GROUP BY
+        tt.line_description, TO_CHAR(l.created_at, 'DD/MM/YYYY'), tt.account_code, cc.cost_centre, tt.fee_type
+	UNION
```
denialanderror, line 38: The two sides of the union are doing exactly the same fairly expensive left joins, which is quite inefficient. It would be better to capture the transactions in the union and then do the left lateral joins over them once unioned.

finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,158 @@
...
+            amount
+        ELSE
+            ''
+        END AS "Credit",
+    line_description || ' [' || transaction_date || ']' AS "Line description"
+FROM
+    partitioned_data
+ORDER BY
+    CASE
+        WHEN line_description LIKE 'AD invoice%' THEN 1
```
denialanderror, line 86: Wouldn't it be easier to do this ordering based on the transaction_type id rather than description? String comparison is expensive.

migrations/20250128161408_transaction_type_journal_download_columns.sql
```diff
@@ -0,0 +1,69 @@
...
+(46, 'Reapply/Reallocate (money to invoice)', true),
+(47, 'BACS Payment', true),
+(48, 'Online card', true);
+
+UPDATE transaction_type
+SET line_description = line_description_update, is_receipt = is_receipt_update
+FROM transaction_type_update
+WHERE transaction_type_id = id;
+
+UPDATE transaction_type SET supervision_level = fee_type WHERE supervision_level = '' AND fee_type IN ('GA', 'GS', 'GT');
```
denialanderror, line 65: If this is being set on the transaction_type, why are the conditions on these in the queries, e.g. non_receipt_transactions line 40?

denialanderror (changes requested):

finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+const NonReceiptTransactionsQuery = `WITH transaction_totals AS (
+   SELECT
+        tt.line_description AS line_description,
+        TO_CHAR(l.created_at, 'DD/MM/YYYY') AS transaction_date,
+        tt.account_code AS account_code,
+        (ABS(SUM(la.amount) / 100.0)::NUMERIC(10, 2))::VARCHAR(255) AS amount,
+		cc.cost_centre,
+		CASE WHEN tt.fee_type IN ('MCR','ZR','ZE','ZH','WO') THEN true ELSE false END AS is_credit
```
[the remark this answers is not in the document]
josephsmith0705: The 'is_receipt' column defines whether the transaction turns up on the receipt or non-receipt journal, whereas this condition defines whether the amount is a 'credit' or 'debit' on the line. I could add this to the transaction_type table, but as it's only being used in this query, for these few few types I figured it wasn't worth bulking up that table with another column. Though happy to do that if you disagree

finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,167 @@
...
+	WHERE i.created_at::DATE = $1
+),
+transaction_totals AS (
+	SELECT
+		tt.line_description,
+		t.created_at AS transaction_date,
+		tt.account_code,
+		ABS(SUM(t.amount) / 100.0)::NUMERIC(10,2)::VARCHAR(255) AS amount,
+		account.cost_centre,
+		CASE WHEN SUM(t.amount) >= 0 THEN true ELSE false END AS is_credit,
```
denialanderror, line 95: The case is redundant.`SUM(t.amount) >= 0` is already a boolean value

finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,167 @@
...
+
+type NonReceiptTransactions struct {
+	Date *shared.Date
+}
+
+const NonReceiptTransactionsQuery = `
+WITH transaction_type_order AS (
+	SELECT 
+		id, 
+		CASE WHEN id = 1 THEN 1
```
denialanderror, line 15: Not sure I like this way of ordering. It feels both fragile (although we are in control of the migrations and ids now, there's no guarantee that will always be the case) and hard to parse (you cannot tell what these refer to without looking in the database). I think it would be better either to go back to ordering by string value and take the performance hit, or put the ordering in the transaction_type table.

finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,167 @@
...
+		WHEN id = 67 THEN 45
+		WHEN id = 68 THEN 46
+		WHEN id = 69 THEN 47
+		ELSE 48
+		END AS index
+	FROM supervision_finance.transaction_type WHERE is_receipt = false
+),
+transactions AS (
+   	SELECT
+		l.created_at::DATE AS created_at,
```
denialanderror, line 68: created_at is redundant as it is always matching to $1

finance-api/internal/db/non_receipt_transactions.go
```diff
@@ -0,0 +1,167 @@
...
+			FROM supervision_finance.invoice_fee_range ifr
+			WHERE ifr.invoice_id = t.invoice_id
+			ORDER BY id DESC
+			LIMIT 1) END AS supervision_level
+	) sl ON TRUE
+	LEFT JOIN LATERAL (
+		SELECT 
+			tto.index, fee_type, account_code, line_description 
+		FROM supervision_finance.transaction_type tt
+		LEFT JOIN transaction_type_order tto ON tt.id = tto.id
```
denialanderror, line 111: Should this be a left join? Presumably if a transaction type doesn't have an order, it shouldn't be in the report?

Same as the is_receipt below. If the transaction type isn't a receipt type, it won't have an order, which means if you aren't left joining on the table, it necessarily won't be returned. Unless I'm misunderstanding something.

finance-api/internal/db/non_receipt_transactions_test.go
denialanderror, line 1: Testing here is still quite light. For instance, it is only testing what will be returned. There's no test data to assert that receipt transactions won't show in this report.

finance-api/internal/db/receipt_transactions.go
```diff
@@ -0,0 +1,98 @@
...
+	WHERE l.created_at::DATE = $1
+	GROUP BY tt.line_description, l.bankdate, tt.account_code, l.type, n, tt.index
+)
+SELECT 	
+	'="0470"'                                              		AS "Entity",
+	'99999999'                                       			AS "Cost Centre",
+	CASE WHEN n % 2 = 1 
+		THEN CASE WHEN ledger_type = 'SUPERVISION BACS PAYMENT' 
+			THEN '1841102088' 
+			ELSE '1841102050'
```
denialanderror, line 54: There's a lot of hardcoded accounting strings like these. The point of adding these to the database was to avoid situations where this is encoded in SQL queries. If we need a receivable account code and a revenue account code, these should probably be added as separate columns to the table.

denialanderror (changes requested):
Comments on the non-receipts journal largely apply also to the receipt one, especially in regard to testing.

denialanderror (approved):

PR #221, opened 2025-02-06
pfs-246 payment method being able to change #minor

Nicholassully:


denialanderror (changes requested):

finance-api/internal/service/service.go
```diff
@@ -17,6 +17,7 @@ type TX interface {
 
 type Dispatch interface {
 	CreditOnAccount(ctx context.Context, event event.CreditOnAccount) error
+	PaymentMethod(ctx context.Context, event event.PaymentMethod) error
```
denialanderror, line 20: This should probably be renamed to match the event name change

Nicholassully: done

PR #226, opened 2025-02-12
PFS-171 Download fee accrual report

josephsmith0705:


finance-api/internal/filestorage/client.go
```diff
@@ -53,22 +53,17 @@ func NewClient(ctx context.Context, region string, iamRole string, endpoint stri
...
-	output, err := c.s3.GetObject(ctx, &s3.GetObjectInput{
-		Key:    &fileName,
-		Bucket: &bucketName,
-	})
-	if err != nil {
-		return nil, err
+func (c *Client) GetFile(ctx context.Context, bucketName string, filename string, versionID string) (*s3.GetObjectOutput, error) {
+	var awsVersionId *string
+
+	if versionID != "" {
```
denialanderror, line 59: Why would the versionID ever be empty? This was added as a security protection, as it is a auto-generated UUID that prevents a user from attempting to access a file by modifying the base64'd JSON.

finance-api/internal/service/process_finance_admin_upload.go
```diff
@@ -16,14 +16,14 @@ import (
 )
 
 func (s *Service) ProcessFinanceAdminUpload(ctx context.Context, detail shared.FinanceAdminUploadEvent) error {
-	file, err := s.fileStorage.GetFile(ctx, os.Getenv("ASYNC_S3_BUCKET"), detail.Filename)
+	file, err := s.fileStorage.GetFile(ctx, os.Getenv("ASYNC_S3_BUCKET"), detail.Filename, "")
```
denialanderror, line 19: I'd rather we had two similar functions with different signatures than one function that accepts empty strings. This runs the risk of a misconfiguration error, where a string value is meant to be passed in but isn't set correctly.

denialanderror (changes requested):

finance-api/internal/filestorage/client.go
josephsmith0705: Scratch that, there's no way to get the version ID without just getting the whole latest file, and it wouldn't really have any security effects anyway. I've set every other usage of this 'getObject' to use the version ID, but this one needs to get the latest one every time

finance-api/internal/reports/generate_report.go
```diff
@@ -28,6 +53,20 @@ func (c *Client) GenerateAndUploadReport(ctx context.Context, reportRequest shar
 		return
 	}
 
+	if reportRequest.ReportType == shared.ReportsTypeAccountsReceivable &&
+		*reportRequest.AccountsReceivableType == shared.AccountsReceivableTypeFeeAccrual {
+		payload, err := c.createDownloadFeeAccrualNotifyPayload(reportRequest.Email, requestedDate)
+		if err != nil {
+			logger.Error("failed to generate notify payload", "error", err)
```
Nicholassully, line 60: Not sure it matters that much but for consistency  logger.Error("failed to generate notify payload", "error", err) be logger.Error("failed to generate notify payload", "err", err)

Nicholassully (dismissed):

Nicholassully (approved):

PR #233, opened 2025-02-19
pfs-216 ingest unapplied payments

Nicholassully:


denialanderror (changes requested):
You have a new billing history event but no corresponding template, so this isn't going to render anything

denialanderror (changes requested):
I'm a bit confused why there is a new billing event where the template is identical to an existing one. Either SOP unallocated events need to be displayed differently, in which case the template should differ from the payment-processed-event, or it doesn't and it should just trigger the existing event.

denialanderror (approved):

PR #234, opened 2025-02-20
PFS-237: created by ids

denialanderror:


finance-api/internal/service/service.go
```diff
@@ -65,3 +67,8 @@ func (s *Service) BeginStoreTx(ctx context.Context) (*store.Tx, error) {
 
 	return store.NewTx(tx), nil
 }
+
+func (*Service) WithCancel(ctx context.Context) (context.Context, context.CancelFunc) {
```
Nicholassully, line 71: Just for my understanding what is this function doing? Is it just in case the transaction fails there is context to delete the request?

Nicholassully (approved):

finance-api/internal/service/service.go
denialanderror: Pretty much. The returned CancelFunc is deferred, so when the context is cancelled, e.g. due to an error, the deferred function is called, which in the case of a DB transaction will roll it back.

This was already being done in the code but it just returns a context.Context, whereas we now need an auth.Context. Therefore, we need to create the cancel context and then add it to our context type. Instead of repeating that pattern everywhere, I created this helper function to abstract that out.

PR #238, opened 2025-02-26
pfs-269 change the migration to check for two columns

Nicholassully:


denialanderror (changes requested):
You cannot alter an existing migration. Not only has this already been run and so won't be picked up, if it was run it would result in undefined behaviour.

PR #256, opened 2025-03-19
Attempt to fix generated reports not uploading

josephsmith0705:


denialanderror (approved):
Would be good to know why this fixed the issue

josephsmith0705:
> Would be good to know why this fixed the issue

That extra defer file.close seems to be closing the file before it can be uploaded, causing this error: "serialization failed: seek: file already closed"

PR #268, opened 2025-03-28
PFS-234: Misapplied payments

denialanderror:


finance-api/internal/service/process_payments.go
```diff
@@ -0,0 +1,260 @@
...
+		return shared.TransactionTypeOPGBACSPayment, nil
+	case shared.ReportTypeUploadPaymentsSupervisionCheque:
+		return shared.TransactionTypeSupervisionChequePayment, nil
+	case shared.ReportTypeUploadDirectDebitsCollections:
+		return shared.TransactionTypeDirectDebitPayment, nil
+	}
+	return shared.TransactionTypeUnknown, fmt.Errorf("unknown upload type")
+}
+
+func parseAmount(amount string) (int32, error) {
```
Nicholassully, line 69: Could you walk me through this function please?

denialanderror: This is the same function that was in process_finance_admin_upload, it has just been moved. It takes string amount as currency and converts it into an integer.

finance-api/cmd/api/process_upload.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	slices.Sort(keys)
+
+	for _, key := range keys {
+		failedLine := failedLines[key]
+		errorMessage = ""
+
+		switch failedLine {
+		case "DATE_PARSE_ERROR":
```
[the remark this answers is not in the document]
denialanderror: They shouldn't be in the shared package because they aren't shared between services, they are only used for the API side. I will create some static values that can be used but we should only be putting things in the shared package that are cross-concerns.

Nicholassully (approved):

PR #292, opened 2025-04-30
pfs-327 adhoc event for neg invoices

Nicholassully:


finance-api/internal/service/process_adhoc_event.go
```diff
@@ -0,0 +1,86 @@
...
+		todaysDate := time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, time.UTC)
+		details := shared.PaymentDetails{
+			Amount: int32(invoice.Ledgerallocationamountneeded),
+			ReceivedDate: pgtype.Timestamp{
+				Time:             todaysDate,
+				InfinityModifier: 0,
+				Valid:            true,
+			},
+			CourtRef:   pgtype.Text{String: invoice.CourtRef.String, Valid: true},
+			LedgerType: shared.TransactionTypeUnappliedPayment,
```
denialanderror, line 38: Unless I am misunderstanding the ticket, "unapplied payment" isn't the ledger type. The ledger type is a payment type and it should be added as an unapply, i.e. with a ledger allocation status of "UNAPPLIED".

finance-api/internal/service/process_adhoc_event.go
```diff
@@ -0,0 +1,86 @@
...
+
+	for _, invoice := range invoices {
+		now := time.Now().UTC()
+		todaysDate := time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, time.UTC)
+		details := shared.PaymentDetails{
+			Amount: int32(invoice.Ledgerallocationamountneeded),
+			ReceivedDate: pgtype.Timestamp{
+				Time:             todaysDate,
+				InfinityModifier: 0,
+				Valid:            true,
```
denialanderror, line 35: We shouldn't be creating pgtypes manually like this. use the `Scan` function for the type to parse the value.

finance-api/cmd/api/process_adhoc_event.go
```diff
@@ -0,0 +1,19 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"github.com/ministryofjustice/opg-sirius-supervision-finance-hub/shared"
+)
+
+func (s *Server) processAdhocEvent(ctx context.Context, event shared.AdhocEvent) error {
+	if event.AdhocProcessName != "NegativeInvoices" {
```
denialanderror, line 10: This part is a bit brittle. Every time we add a new adhoc event, we will need to change tests. It would probably be better if the valid events were injected into the `Server` on construction. That way, it can be tested, agnostic of the actual events being parsed.

shared/event.go
```diff
@@ -76,3 +84,7 @@ type FinanceAdminUploadEvent struct {
 	UploadDate   Date             `json:"uploadDate"`
 	PisNumber    int              `json:"pisNumber"`
 }
+
+type AdhocEvent struct {
+	AdhocProcessName string `json:"adhocProcessName"`
```
denialanderror, line 89: `adhocProcessName` feels verbose, seeing as we already know it is an adhoc process. `name` or `task` would be better.

shared/report_upload_type.go
```diff
@@ -44,6 +46,7 @@ var reportTypeUploadMap = map[string]ReportUploadType{
 	"DEPUTY_SCHEDULE":             ReportTypeUploadDeputySchedule,
 	"DIRECT_DEBITS_COLLECTIONS":   ReportTypeUploadDirectDebitsCollections,
 	"MISAPPLIED_PAYMENTS":         ReportTypeUploadMisappliedPayments,
+	"UPLOAD_ADHOC":                ReportTypeUploadAdhoc,
```
denialanderror, line 49: Is this left over from a previous iteration? I don't see a report being uploaded as part of this process.

denialanderror (dismissed):

denialanderror:
It would be good to include an ADR for this to detail the decision to use eventbridge for adhoc processes

shared/transaction_type.go
```diff
@@ -39,6 +40,7 @@ var TransactionTypeMap = map[string]TransactionType{
 	"DIRECT DEBIT PAYMENT":       TransactionTypeDirectDebitPayment,
 	"SUPERVISION CHEQUE PAYMENT": TransactionTypeSupervisionChequePayment,
 	"SOP UNALLOCATED":            TransactionTypeSOPUnallocatedPayment,
+	"UNAPPLIED PAYMENT":          TransactionTypeUnappliedPayment,
```
denialanderror, line 43: This can be removed.

finance-api/internal/service/process_adhoc_event.go
```diff
@@ -0,0 +1,104 @@
...
+
+		if invoice.Type == "UNKNOWN CREDIT" {
+			if invoice.Ledgerid == 652949 || invoice.Ledgerid == 652948 {
+				invoice.Type = "CREDIT REMISSION"
+			} else {
+				invoice.Type = "CREDIT EXEMPTION"
+			}
+		}
+
+		fmt.Print(invoice.Type)
```
denialanderror, line 42: Fine to log this if needed but use the logger otherwise it won't appear in the console. Otherwise remove.

denialanderror (changes requested):

denialanderror (approved):

PR #297, opened 2025-05-08
pfs-325 moving sqlc to docker

Nicholassully:


Makefile
```diff
@@ -53,7 +53,7 @@ clean:
 	docker compose down
 	docker compose run --rm yarn
 
-up: clean build-dev start-and-seed sqlc-gen
```
sixdaysandy, line 56: Doesn't this still want to be sqlc-gen as that's the command name?

Nicholassully: done

sixdaysandy (approved):

PR #310, opened 2025-05-15
pfs-333 adding an ignore

Nicholassully:


zap/scan.sh
```diff
@@ -84,7 +84,8 @@ filter_false_positives() {
     # This will globally report all alert results of this type as a false positive and exclude them from the reports
     # These Alert ID's can be found by viewing the ids of the /JSON/pscan/view/scanners/ and /JSON/ascan/view/scanners/ endpoints
     ALERTS=(
-        "10031" # User Controllable HTML Element Attribute (Potential XSS)
+        "10031", # User Controllable HTML Element Attribute (Potential XSS)
+        "6"
```
denialanderror, line 88: Can you add a comment with the alert name like above please?

denialanderror (changes requested):

kate-49 (approved):

PR #321, opened 2025-05-30
pfs-360: Floating point error in decimal string translation resulting…

denialanderror:
… in some invoices being some amounts created a penny out


shared/transform.go
```diff
@@ -1,21 +1,33 @@
...
+	if len(parts) == 1 {
+		parts = append(parts, "00")
+	} else if len(parts[1]) == 1 {
+		parts[1] += "0"
+	} else if len(parts[1]) > 2 {
+		parts[1] = parts[1][:2] // truncate extra decimals
+	}
+
+	combined := parts[0] + parts[1]
+	val, _ := strconv.ParseInt(combined, 10, 32)
```
mohammedrmoj, line 31: @denialanderror What would happen if `s` is an invalid string like for example "1a.2b" ?

denialanderror: It would panic, but in reality this can't happen as the string is coming from a numeric form field, so non-numeric characters won't be present.

PR #323, opened 2025-06-02
PFS-350 block duplicate reversals

josephsmith0705:


finance-api/internal/service/process_reversals.go
```diff
@@ -184,29 +188,48 @@ func getReversalLines(ctx context.Context, record []string, uploadType shared.Re
...
+		ReceivedDate: details.ReceivedDate,
+		PisNumber:    details.PisNumber,
+	})
+
+	if reversalCount >= ledgerCount {
+		(*failedLines)[index] = validation.UploadErrorDuplicateReversal
+		return false
+	}
+
+	if fileReversalCount(processedRecords, details) >= int(ledgerCount) {
```
denialanderror, line 228: I'm not quite sure I understand what is happening here. This function is validating a single reversal line against the data in the DB, so I'm not clear on why we would need to also check it against the processed records. Won't they be in the database if they have been processed?

finance-api/internal/service/process_reversals.go
```diff
@@ -283,3 +306,13 @@ func (s *Service) ProcessReversalUploadLine(ctx context.Context, tx *store.Tx, d
 
 	return nil
 }
+
+func fileReversalCount(processedRecords []shared.ReversalDetails, details shared.ReversalDetails) int {
```
denialanderror, line 310: This name is a bit confusing. Maybe `totalReversedPayments`? Or what might be better in terms of making the code self-documenting is to move the condition into the function, e.g. 
```func hasPaymentToReverse(processedRecords []shared.ReversalDetails, details shared.ReversalDetails, totalPayments int) bool {
    reversals := 0
	for _, s := range processedRecords {
		if s == details {
			reversals++
		}
	}
	return reversals > totalPayments
}
```

denialanderror (approved):

PR #343, opened 2025-06-26
PFS-243 Upload debt chase report

josephsmith0705:


finance-api/cmd/api/process_upload.go
```diff
@@ -69,47 +70,57 @@ func (s *Server) processUpload(w http.ResponseWriter, r *http.Request) error {
...
 	logger := s.Logger(ctx)
 
-	csvReader := csv.NewReader(upload.FileBytes)
-	records, err := csvReader.ReadAll()
-	if err != nil {
-		logger.Error("unable to read report", "err", err)
-		payload := createUploadNotifyPayload(upload.EmailAddress, upload.UploadType, fmt.Errorf("unable to read report"), map[int]string{})
-		err = s.notify.Send(ctx, payload)
+	if upload.UploadType.IsDirectUpload() {
+		err := s.service.ProcessDirectUploadReport(ctx, fmt.Sprintf("fee-chase/%s.csv", strings.ToLower(upload.UploadType.Key())), upload.FileBytes)
```
denialanderror, line 77: Is it necessary to pass the filename here when it is composed by from the upload type? It would be better if the upload type was passed in and then the filename was composed in the function, especially as here it is hardcoded to "fee-chase/"

denialanderror (changes requested):

denialanderror (approved):

denialanderror (approved):

PR #351, opened 2025-07-02
PFS-366 Sort invoices by raised date

josephsmith0705:


denialanderror (approved):
It may be useful to add some comments or documentation to this to explain why we need to track index as well as ID

PR #371, opened 2025-07-22
SFS-5 cancel direct debit page

josephsmith0705:


finance-hub/web/template/layout/person-info.gotmpl
```diff
@@ -15,15 +15,27 @@
     </div>
     {{ if .User.IsFinanceUser }}
             {{ if eq showDirectDebits true }}
-                 <a class="govuk-button moj-button-menu__item govuk-button--secondary govuk-!-margin-top-5"
+                {{ if eq .FinanceClient.PaymentMethod "Demanded" }}
```
kate-49, line 18: What does this default to if it's not set? Just checking it won't freak out if it tries to access paymentMethod and it's null.

josephsmith0705: If it's not set, no button will be shown either to set up a direct debit or cancel a direct debit - however, this column is constrained as 'not null' in the database, so if it's not set then something's gone wrong anyway

kate-49 (approved):

PR #376, opened 2025-07-29
SFS-145: Create DD mandate

denialanderror:


mohammedrmoj (approved):

finance-hub/internal/allpay/create_mandate.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Address struct {
+	Line1    string `json:"Line1"`
+	Town     string `json:"Town"`
+	PostCode string `json:"PostCode"`
+}
+
+type CreateMandateRequest struct {
+	SchemeCode      string  `json:"schemeCode"`
+	ClientReference string  `json:"ClientReference"`
```
kate-49, line 26: schemeCode is starting with a lower case while everything else is capitalised, is this a typo?

json-server/config/db.json
```diff
@@ -44,109 +44,393 @@
...
+        "deputyStatus": "Active"
+      },
+      "activeCaseType": {
+        "handle": "HW",
+        "label": "Health & Welfare"
+      },
+      "clientStatus": {
+        "handle": "ACTIVE",
+        "label": "Active"
+      }
```
kate-49, line 63: Not required but all these active case types and statuses are the same type. Is there value in having some as Pfa/ other client statuses for tests?

adrs/00027-client-data-dd.md
```diff
@@ -0,0 +1,35 @@
...
+## Consequences
+
+Fetching the data in `finance-hub` means the front end service will be responsible for validation, whereas this is usually
+done in `finance-api`. This should not cause issues but is a departure from established practice.
+
+The clients endpoint in the Sirius API is very bulky and contains a lot of data that isn't relevant, e.g. every task ID
+associated with the client. This is likely because it is used by multiple external services, as well as being used as the
+internal API for the Sirius UI. This conflict could make the API very brittle, as a change for one service could result 
+in other services breaking. As a result, we should introduce contract testing with Pact to ensure the API we are consuming
+adheres to its contract.
```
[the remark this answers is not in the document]
denialanderror: This was considered and may be a route we go down if Pact is too onerous.

json-server/config/db.json
denialanderror: Not at the moment. The fail case test already tests on fee payer not being set, so the rest of the data is just there for completion.

PR #379, opened 2025-07-31
SFS-22 Update payment processed event and used for all payment types

josephsmith0705:


finance-api/internal/service/get_billing_history.go
```diff
@@ -334,31 +334,43 @@ func processLedgerAllocations(allocations []store.GetLedgerAllocationsForClientR
...
 				event.Amount += int(allocation.AllocationAmount)
+			} else if event.TransactionType.IsPayment() {
+				event.Amount -= int(allocation.AllocationAmount)
 			}
 
 			lh = &historyHolder{
 				billingHistory: shared.BillingHistory{
 					User:  int(allocation.CreatedBy.Int32),
-					Date:  shared.Date{Time: allocation.CreatedAt.Time},
+					Date:  shared.Date{Time: allocation.LedgerDatetime.Time},
```
josephsmith0705, line 357: Confusingly, this 'CreatedAt' value was fetching the 'datetime' in the ledger table. I've updated the query to fetch the actual 'created_at' column in place of this, and updated the 'datetime' column to be fetched with this new name, LedgerDatetime

finance-api/internal/service/get_billing_history_test.go
```diff
@@ -951,7 +953,7 @@ func Test_processLedgerAllocations(t *testing.T) {
 					Status:           "REAPPLIED",
 					LedgerAmount:     5000,
 					AllocationAmount: 5000,
-					CreatedAt: pgtype.Timestamp{
+					LedgerDatetime: pgtype.Timestamp{
```
kate-49, line 956: Just checking because it caught me out in refunds. Do we expected an unapplied/ reapplied timeline event to also have an allocated timeline event? If so it'd be good to test for both of these on one case.

shared/transaction_type.go
```diff
@@ -121,6 +133,10 @@ func (t TransactionType) Key() string {
 	}
 }
 
+func (u TransactionType) IsPayment() bool {
+	return slices.Contains(paymentTransactionTypes, u)
+}
+
```
kate-49, line 139: Might be good to have a test for this one, just in case we change payment Transaction Types at any point without realising the implication?

finance-hub/web/template/billinghistory/payment-processed-event.gotmpl
```diff
@@ -15,7 +25,19 @@
...
+                                {{ printf "£%v unallocated" (toCurrency .Amount) }}
+                            {{ end }}
+                        {{ else }}
+                            {{if eq .Status "ALLOCATED" }}
+                                {{ printf "£%v reversed against %v" (toCurrency .Amount) .InvoiceReference.Reference }}
+                            {{ else }}
+                                {{ printf "£%v reversed on account balance" (toCurrency .Amount) }}
+                            {{ end }}
+                        {{ end }}
                     </li>
```
kate-49, line 41: Would we normally do an e2e test for this?

finance-api/internal/service/get_billing_history_test.go
josephsmith0705: If I'm understanding you correctly, to my knowledge the unapplies and reapplies are instantly allocated so that would mean we have 2 functionally identical timeline events each time an unapply or reapply is done

PR #426, opened 2025-10-02
sfs-236 fix payment with multiple allocations

kate-49:


cypress/e2e/billing-history.cy.js
```diff
@@ -29,7 +29,7 @@ describe("Billing History Tab", () => {
         });
 
         cy.get(".moj-timeline__item").eq(4).within(() => {
-            cy.get(".moj-timeline__title").contains("MOTO card payment of £100 received");
+            cy.get(".moj-timeline__title").contains("MOTO card payment of £130 received");
```
kate-49, line 32: This is now accurate, you can see the two lines below reference a credit balance of £30 and another of £100

PR #430, opened 2025-10-06
Sfs 55 cancel dd close pending

denialanderror:


api-mocks/allpay/allpay-config.yml
```diff
@@ -56,13 +56,13 @@ resources:
       content: "Bad Request - Please check the format of your body request"
 
   - path: "/AllpayApi/Customers/{SchemeCode}/{ClientRef}/{Surname}/Mandates/{Date}"
-    method: post
+    method: delete
     response:
       statusCode: 200
       file: data/success.json
```
kate-49, line 62: Just wanted to check this status code is correct, obviously its an allpay decision but i'd expect a 204

api-mocks/allpay/allpay-config.yml
```diff
@@ -71,6 +71,21 @@ resources:
       statusCode: 400
       content: "Bad Request - Please check the format of your body request"
 
+  - path: "/AllpayApi/Customers/{SchemeCode}/{ClientRef}/{Surname}/Mandates/{Date}/{Amount}"
+    method: delete
+    response:
+      statusCode: 200
+      file: data/success.json
```
kate-49, line 78: Same comment about is a delete really a 200

kate-49 (dismissed):

api-mocks/allpay/allpay-config.yml
denialanderror: Yes, all valid Allpay responses (according to their docs) return 200

api-mocks/allpay/allpay-config.yml
```diff
@@ -71,6 +71,21 @@ resources:
...
+
+  - path: "/AllpayApi/Customers/{SchemeCode}/{ClientRef}/{Surname}/Mandates/{Date}/{Amount}"
+    method: delete
+    pathParams:
+      # allpay_remove_schedule_validation
+      Surname: YWxscGF5X3JlbW92ZV9zY2hlZHVsZV92YWxpZGF0aW9u
```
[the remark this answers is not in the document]
denialanderror: Yes, otherwise you would need to manually decode the base64 encoded string to find out what surname to use

PR #548, opened 2026-04-13
sfs-484: Ensures cancel mandate request is only sent to allpay if the…

denialanderror:
…re is a mandate to cancel

### DIFF
--- a/finance-api/internal/service/cancel_dd_mandate.go
+++ b/finance-api/internal/service/cancel_dd_mandate.go
@@ -15,13 +15,6 @@ import (
 )
 
 func (s *Service) CancelDirectDebitMandate(ctx context.Context, clientID int32, cancelMandate shared.CancelMandate) error {
-	logger := s.Logger(ctx)
-
-	if !s.env.AllpayEnabled {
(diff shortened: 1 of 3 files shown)


finance-api/internal/service/cancel_dd_mandate.go
```diff
@@ -15,6 +15,14 @@ import (
 )
 
 func (s *Service) CancelDirectDebitMandate(ctx context.Context, clientID int32, cancelMandate shared.CancelMandate) error {
+	client, err := s.store.GetClientById(ctx, clientID)
+	if err != nil {
+		return err
+	}
+	if client.PaymentMethod != shared.PaymentMethodDirectDebit.Key() {
```
MiaGordon91, line 22: Just for my understanding, we skip cancelation requests if the clients payment method is not direct debit - is that correct?

denialanderror: Yes, if the client isn't paying by direct debit, there isn't a mandate to cancel. we only change the payment method to demanded after the mandate has been cancelled, so at this point it is still set to direct debit.

MiaGordon91 (approved):

PR #570, opened 2026-05-06
Sfs 560 remove schedules

denialanderror:
We need to remove future payment schedules for every client that were set up as placeholders but due to some miscommunication are still present. This requires approximately 13,000 API calls to Allpay and as there are no guarantees of response times, it is not possible to do this in batches or as a long running job.

The solution is to use the existing Eventbridge pattern to queue the removals so they can each happen asynchronously and independent of each other.

The file gets uploaded to Finance Admin in the same way as payment files with the date of collection. This will require multiple uploads, as the collections are spread over many months but this is probably desired, as it means we can batch and monitor the removals more effectively. The file is then sent to Finance API, which parses the file and dispatches "schedule-to-remove" event. EventBridge then sends this event back to Finance API, which sends the removal request to Allpay.


finance-api/cmd/api/process_upload.go
```diff
@@ -126,6 +126,12 @@ func (s *Server) processUploadFile(ctx context.Context, upload Upload) {
 				logger.Error(fmt.Sprintf("unable to process reverse refunds due to %d failed lines", len(failedLines)))
 			}
 			payload = createUploadNotifyPayload(upload.EmailAddress, upload.UploadType, err, failedLines)
+		} else if upload.UploadType.IsScheduleRemoval() {
+			failedLines := s.service.QueueScheduleRemovals(ctx, records, upload.UploadDate)
+			if len(failedLines) > 0 {
+				logger.Error(fmt.Sprintf("unable to process schedule removal due to %d failed lines", len(failedLines)))
+			}
```
kate-49, line 133: Does this cause an error in the infra alarms?

kate-49 (dismissed):

kate-49:
Just wondering if any of these direct debits have been collected yet?

finance-api/cmd/api/process_upload.go
denialanderror: No. We wouldn't want it to as the user is informed of the issue via notify and would be dealing with it themselves.

denialanderror:
> Just wondering if any of these direct debits have been collected yet?

No. The earliest collection is 1st June

kate-49 (approved):

PR #600, opened 2026-07-07
sfs-641 create adhoc task to fetch mandate and schedule data

MiaGordon91:
Database query retrieves clients that have had a direct debit payment method set on or after the 28 April.
Two calls are made to the Allpay API to fetch mandate and schedule info for each court ref and a csv is created containing those that have a missing schedule.


api-mocks/allpay/allpay-config.yml
```diff
@@ -39,6 +39,46 @@ resources:
       file: data/success.json
       template: true
 
+  - path: "/AllpayApi/Customers/{SchemeCode}/{ClientRef}/{Surname}"
+    method: get
+    pathParams:
+      # finance_client.court_ref/surname: 27272700 / schedule_validation - mandate with schedule
```
josephsmith0705, line 45: Do we need these comments here?

finance-api/internal/allpay/dd_mandate_schedule_check.go
```diff
@@ -0,0 +1,102 @@
...
+	mandate, mandateErr := c.fetchMandateData(ctx, data)
+	schedule, scheduleErr := c.fetchScheduleData(ctx, data)
+
+	// Return a combined result with any errors from the individual fetches. This allows the caller to handle partial failures gracefully.
+	output := &MandateScheduleCheckOutput{
+		Mandate:  mandate,
+		Schedule: schedule,
+	}
+
+	// if mandate/schedule fails it still produces usable result and won't stop the command on the first bad Allpay response
```
josephsmith0705: As above, I'm not sure if comments in this file are necessary as the code seems to be self-documenting enough imo :)

finance-api/internal/service/dd_mandate_schedule_check.go
josephsmith0705, line 1: Could we have a test for this file to ensure the allpay API is called as we expect, the CSV is written and uploaded, etc?

api-mocks/allpay/allpay-config.yml
denialanderror: I'd keep these. The path params are base64 encoded, so without the comments you would need to decode each time you wanted to work out which client they referred to.

finance-api/internal/allpay/dd_mandate_schedule_check.go
```diff
@@ -0,0 +1,102 @@
...
+
+type MandateScheduleCheckOutput struct {
+	Mandate       *FetchMandateScheduleOutput
+	Schedule      *FetchMandateScheduleOutput
+	MandateError  string
+	ScheduleError string
+}
+
+// FetchMandateSchedule fetches mandate and schedule data for one client so the caller can write a single CSV row.
+func (c *Client) FetchMandateSchedule(ctx context.Context, data FetchMandateScheduleInput) (*MandateScheduleCheckOutput, error) {
```
denialanderror: I'd prefer if this was broken out into two separate functions on the Allpay client, e.g. FetchMandate and FetchSchedules. The intention behind the client is that is the access point to the API, so it shouldn't do more work than it needs to, and it should be agnostic as to why it is called, e.g. it shouldn't know it is fetching data for a CSV.

The logic to call the two endpoints and combine the data should be in the service layer.

denialanderror (changes requested):

finance-api/internal/service/dd_mandate_schedule_check.go
```diff
@@ -0,0 +1,156 @@
...
+		if err = writeCSVRow(writer, client.CourtRef.String, client.Surname.String, result); err != nil {
+			return err
+		}
+
+		if (i+1)%50 == 0 || i+1 == len(clients) {
+			logger.Info("DD mandate & schedule check: progress", "processed", i+1, "total", len(clients))
+		}
+
+		if i+1 < len(clients) {
+			time.Sleep(ddMandateScheduleCheckSleep)
```
denialanderror, line 79: What's the sleep for?

finance-api/internal/service/dd_mandate_schedule_check.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/ministryofjustice/opg-sirius-supervision-finance-hub/finance-api/internal/allpay"
+)
+
+const (
+	ddMandateScheduleCheckFromDate = "2026-04-28"
+	ddMandateScheduleCheckSleep    = 250 * time.Millisecond
+)
+
+func (s *Service) CheckDirectDebitMandateSchedule(ctx context.Context, logger *slog.Logger) error {
```
denialanderror: The naming of this implies it checks one direct debt schedule, whereas it generates a report of all the clients that don't have schedules set correctly.

finance-api/internal/allpay/dd_mandate_schedule_check.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+type MandateScheduleCheckOutput struct {
+	Mandate       *FetchMandateScheduleOutput
+	Schedule      *FetchMandateScheduleOutput
+	MandateError  string
+	ScheduleError string
+}
+
+func (c *Client) FetchMandate(ctx context.Context, data FetchMandateScheduleInput) (*FetchMandateScheduleOutput, error) {
```
denialanderror: The convention in this project is each client function exposed to the interface has it's own file. The way you've done it does reduce some boilerplate and may be a good option for the future, but for maintainability, it would be better if it followed the other projects and each function handles its own API call and error handling.

denialanderror (changes requested):

finance-api/internal/service/dd_mandate_schedule_check.go
MiaGordon91: Added a sleep in after each client to avoid overwhelming Allpay as it sends two API calls for over 1,500 clients but I can remove it if you think it's not needed

denialanderror (approved):

finance-api/internal/service/dd_mandate_schedule_check.go
denialanderror: I'd hope it wouldn't be, as it is just GET requests and we had much larger volumes of POSTs during annual billing, but seeing as Allpay themselves still don't seem to understand what their rate limits are, it is probably worth keeping!

PR #601, opened 2026-07-08
sfs-629: Changes mandate and schedule creation to combine into one AP…

denialanderror:
…I call to avoid issues where the schedule call fails after a successful mandate creation.


MiaGordon91 (commented):
Just for my understanding, would this refactor have any impact on instances where Allpay receive duplicate requests that result in duplicate schedules existing in Allpay?

denialanderror:
> Just for my understanding, would this refactor have any impact on instances where Allpay receive duplicate requests that result in duplicate schedules existing in Allpay?

@MiaGordon91 It shouldn't have any impact. The duplicate schedules are due to the invoice created events from Eventbridge, as SQS is "at-least-once" delivery, so an event may be processed multiple times. This change is triggered when a user presses a button, so it would only send once.

MiaGordon91 (approved):

